PR #66, opened 2026-01-23
Implement `hasattr` builtin function

friendlymatthew:
This PR implements `hasattr(object,name)` which returns `True` if the object has the named attribute, `False` otherwise. Unlike #65, this function never raises an AttributeError

This implementation calls `py_get_attr()` directly and converts the result to a boolean rather than reusing `builtin_getattr()`, since the functions have different signatures to validate


davidhewitt (approved):
Thanks!

PR #68, opened 2026-01-23
Implement `del` statement for subscript deletion

friendlymatthew:
This PR adds support for the `del` statement on subscripted objects, which enable deletion of list elements and dict entries through the `del foo['bar']` syntax


davidhewitt:
Will close in favour of #309

PR #89, opened 2026-02-02
don't leak in `id()`

davidhewitt:
The justification in `id` to leak seems completely baseless, as CPython indeed returns `True` in the case mentioned in the comment:

```python
$ uvx python
Python 3.14.2 free-threading build (main, Dec  9 2025, 20:27:05) [Clang 21.1.4 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> id([]) == id([])
True
```

There is one test failure, for `id([]) != id(())`, the way to solve this IMO would be to make `()` a singleton like it is in CPython (which would be a performance win anyway).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -178,7 +178,7 @@ jobs:
           args: --out dist
           rust-toolchain: stable
           working-directory: crates/monty-python
-      - uses: actions/upload-artifact@v4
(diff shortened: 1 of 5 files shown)


samuelcolvin:
okay, I'm good with this change.

But let's add those cases I asked about to tests with an explanation to make the behaviour explicit.

PR #114, opened 2026-02-06
Fix ast depth overflow

### LINKED ISSUE #112: Nested parentheses parsing can segfault
```python
import argparse
import pydantic_monty

parser = argparse.ArgumentParser()
parser.add_argument("--depth", type=int, default=5750)
args = parser.parse_args()

target = "x"
for _ in range(args.depth):
    target = f"({target},)"

code = f"{target} = (1,)"
print(f"About to construct Monty with nested-unpack depth={args.depth}")
pydantic_monty.Monty(code)

# If vulnerable threshold is not reached, this line prints.
print("No crash at this depth")
```

(may need to tweak the value on different hosts to cause the crash)

samuelcolvin:
fix #112

### DIFF
--- a/crates/monty/tests/parse_errors.rs
+++ b/crates/monty/tests/parse_errors.rs
@@ -318,3 +318,82 @@ fn deeply_nested_attribute_access_exceed_limit() {
     let result = MontyRun::new(code, "test.py", vec![], vec![]);
     assert_eq!(get_exc_type(result), ExcType::SyntaxError);
 }
+
+#[test]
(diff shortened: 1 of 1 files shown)


samuelcolvin:
@ankrgyl I think the fix is complete. Would you like to try this branch, or shall I merge and release it?

ankrgyl:
please feel free to proceed without me — am preoccupied for a bit. would be happy to test it again over the weekend though

PR #121, opened 2026-02-07
Guard for recursive opperations

### LINKED ISSUE #116: Recursive object construction can segfault
While I was poking around, I found another segfault:

```
import argparse
import pydantic_monty


def build_code(depth: int) -> str:
    return f"x=[]\nfor _ in range({depth}):\n    x=[x]\nx"


def main() -> None:
    parser = argparse.ArgumentParser()
    parser.add_argument("--depth", type=int, default=1200)
    args = parser.parse_args()

    code = build_code(args.depth)
    print(f"building nested runtime value depth={args.depth}", flush=True)

    monty = pydantic_monty.Monty(code)
    print("constructed Monty; running now", flush=True)

    try:
        out = monty.run()
        print("no crash at this depth; type:", type(out), flush=True)
    except BaseException as exc:
        print("caught exception:", type(exc).__name__, exc, flush=True)


if __name__ == "__main__":
    main()
```

Running this with depth 10000 on my machine segfaults.

samuelcolvin:
Fix #116.

### DIFF
--- a/crates/monty/src/heap.rs
+++ b/crates/monty/src/heap.rs
@@ -244,7 +244,7 @@ impl HeapData {
             }
             Self::Tuple(t) => {
                 // Guard against deep nesting - if limit exceeded, treat as unhashable
-                if heap.try_inc_data_recursion().is_err() {
+                if heap.increase_data_recursion().is_err() {
                     return None;
                 }
                 let mut hasher = DefaultHasher::new();
@@ -253,17 +253,17 @@ impl HeapData {
                 for obj in t.as_vec() {
                     let h = obj.py_hash(heap, interns);
                     if h.is_none() {
-                        heap.dec_data_recursion();
+                        heap.reduce_data_recursion();
                         return None;
                     }
                     h.unwrap().hash(&mut hasher);
                 }
-                heap.dec_data_recursion();
+                heap.reduce_data_recursion();
                 Some(hasher.finish())
             }
             Self::NamedTuple(nt) => {
                 // Guard against deep nesting - if limit exceeded, treat as unhashable
-                if heap.try_inc_data_recursion().is_err() {
+                if heap.increase_data_recursion().is_err() {
                     return None;
                 }
                 let mut hasher = DefaultHasher::new();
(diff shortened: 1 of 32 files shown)


davidhewitt (commented):
~~I would think a recursive function, e.g.~~

```python
def foo():
    foo()
```

~~is probably similarly problematic?~~

EDIT: I see that's already covered previously.

crates/monty/src/heap.rs
```diff
@@ -856,6 +910,10 @@ pub(crate) struct Heap<T: ResourceTracker> {
     may_have_cycles: bool,
     /// Number of GC applicable allocations since the last GC.
     allocations_since_gc: u32,
+    /// Tracks recursion depth for container operations (repr, eq, cmp, hash).
+    /// Uses AtomicU16 for thread safety with Python bindings.
+    /// Prevents stack overflow from deeply nested (but not cyclic) structures.
+    data_recursion_depth: AtomicU16,
```
davidhewitt: I feel like the `AtomicU16` must be a complete hallucination, it seems to me like pretty much every operation uses `&mut Heap` so there should never be a need for an atomic here.

crates/monty/src/heap.rs
```diff
@@ -1068,34 +1219,40 @@ impl<T: ResourceTracker> Heap<T> {
...
     /// Panics if the value ID is invalid or the value has already been freed.
     pub fn dec_ref(&mut self, id: HeapId) {
-        let slot = self.entries.get_mut(id.index()).expect("Heap::dec_ref: slot missing");
-        let entry = slot.as_mut().expect("Heap::dec_ref: object already freed");
-        if entry.refcount > 1 {
-            entry.refcount -= 1;
-        } else if let Some(value) = slot.take() {
-            // refcount == 1, free the value and add slot to free list for reuse
-            self.free_list.push(id);
+        let mut work_list = vec![id];
```
davidhewitt: Should avoid allocating, e.g. by using a `smallvec`, maybe also having 

```rust
let mut current = id;
let mut additional = Vec::new();

loop {
    // do stuff with current
    
    let Some(next) = additional.pop() else { break };
    current = next;
}
```

99991:
In CPython:

```python
x = []
x.append(x)
print(repr(x)) # prints [[...]] with "..." instead of recursion.
```

Why do you raise an exception instead?

PR #122, opened 2026-02-07
always allocate empty tuple singleton at index 0

davidhewitt:
Hoping this might be even better for perf, let's see...

### DIFF
--- a/crates/monty/src/heap.rs
+++ b/crates/monty/src/heap.rs
@@ -944,7 +944,7 @@ impl<T: ResourceTracker> Heap<T> {
         let empty_tuple = this
             .allocate(HeapData::Tuple(Tuple::default()))
             .expect("Failed to allocate empty tuple singleton");
-        assert_eq!(empty_tuple, EMPTY_TUPLE_ID);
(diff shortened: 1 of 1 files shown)


crates/monty/src/heap.rs
```diff
@@ -938,14 +932,20 @@ impl<T: ResourceTracker> Heap<T> {
...
             allocations_since_gc: 0,
-            empty_tuple_id: None,
-        }
+        };
+        // TBC: should the empty tuple contribute to the resource limits?
+        // If not, can just place it in `entries` directly without going through `allocate()`.
+        let empty_tuple = this
+            .allocate(HeapData::Tuple(Tuple::default()))
+            .expect("Failed to allocate empty tuple singleton");
+        assert_eq!(empty_tuple, EMPTY_TUPLE_ID);
```
samuelcolvin: Debug assert?

davidhewitt:
```suggestion
        debug_assert_eq!(empty_tuple, EMPTY_TUPLE_ID);
```

PR #126, opened 2026-02-09
Add suspendable REPL execution

ex3ndr:
Hello!

I will be straight - it was generated by codex, i reviewed it and seemed fine - it clones compiled functions, internal states but keeps heap untouched.

This is a pretty much a must for RLM implementations which monty is perfect for.

### DIFF
--- a/crates/monty-js/wrapper.ts
+++ b/crates/monty-js/wrapper.ts
@@ -20,7 +20,6 @@ import {
   MontySnapshot as NativeMontySnapshot,
   MontyComplete as NativeMontyComplete,
   MontyException as NativeMontyException,
-  MontySnapshot as NativeMontySnapshot,
   MontyTypingError as NativeMontyTypingError,
 } from './index.js'
 


crates/monty/src/run.rs
```diff
@@ -168,6 +230,167 @@ impl MontyRun {
     }
 }
 
+/// Stateful REPL session that executes snippets incrementally without replay.
```
samuelcolvin: this all needs to go in a new file.

crates/monty/src/run.rs
```diff
@@ -299,6 +522,304 @@ impl<T: ResourceTracker + serde::de::DeserializeOwned> RunProgress<T> {
     }
 }
 
+/// Result of a single suspendable REPL snippet execution.
```
samuelcolvin: same.

samuelcolvin:
Also required (could be done in this PR or a followup):
* add support to `monty-cli`
* add support to python
* add support to js

ex3ndr:
Updated:
1) move repl out of run.rs - latter now matches the origin completely
2) added js, cli and python implementations
3) fixed other comments

crates/monty-cli/src/main.rs
```diff
@@ -10,7 +15,25 @@ const EXT_FUNCTIONS: bool = false;
 
 fn main() -> ExitCode {
     let args: Vec<String> = env::args().collect();
-    let file_path = if args.len() > 1 { &args[1] } else { "example.py" };
+    let repl_mode = matches!(args.get(1).map(String::as_str), Some("--repl" | "-r"));
```
davidhewitt: This feels brittle, I would recommend we convert argument parsing to `clap`. (Might be better to do that as a separate precursor PR.)

davidhewitt (commented):
This is a huge PR, I will probably review in multiple rounds. I have some initial questions which might affect the content here, so let's address those first before I review further.

ex3ndr:
Resolved all comments, except arg parsing, but it is robust.

PR #136, opened 2026-02-10
Add comparison to str

lsbrillant:
Add str comparison support.
bringing higher feature parity with base python.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -33,6 +33,7 @@
       "Bash(gh issue view:*)",
       "Skill(python-playground)",
       "Skill(fastmod)",
+      "Skill(coverage)",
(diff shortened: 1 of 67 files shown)


crates/monty/test_cases/str__ops.py
```diff
@@ -132,3 +132,8 @@
 s = 'abc'
 assert s[False] == 'a', 'str getitem with False'
 assert s[True] == 'b', 'str getitem with True'
+
+# Sorting and comparisons
```
samuelcolvin: we should add a test for non-ascii strings and strings of different lengths - I have no idea if python's string comparison logic is the same as Rust's

samuelcolvin (commented):
otherwise LGTM

PR #146, opened 2026-02-11
Add printer callback for javascript binding

### LINKED ISSUE #110: stdout / stderr is not accessible in the JS library
It looks like the stdout is not accessible in the JS bindings. For instance if I have the following executed Python code:

```
print("This goes to stdout")
"This is the return value"
```

After the run the MontyComplete object will only return in the output: `"This is the return value"` but not stdout.

Embers-of-the-Fire:
This PR adds a `printerCallback` field for `runOptions` in `monty-js`.

Should close #110

### DIFF
--- a/crates/monty-js/src/monty_cls.rs
+++ b/crates/monty-js/src/monty_cls.rs
@@ -774,9 +774,9 @@ impl MontyComplete {
 type JsPrintCallback<'env> = Function<'env, FnArgs<(&'static str, String)>, ()>;
 type JsPrintCallbackRef = FunctionRef<FnArgs<(&'static str, String)>, ()>;
 
-/// A `PrintWriter` implementation that calls a Python callback for each print output.
+/// A `PrintWriter` implementation that calls a javascript callback for each print output.
 ///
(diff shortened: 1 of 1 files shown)


crates/monty-js/src/monty_cls.rs
```diff
@@ -712,35 +770,39 @@ impl MontyComplete {
     }
 }
 
+// Function type for JS callback used in `CallbackStringPrint`.
+type JsCallbackFunction<'env> = Function<'env, FnArgs<(&'static str, String)>, ()>;
+type JsCallbackFunctionRef = FunctionRef<FnArgs<(&'static str, String)>, ()>;
```
davidhewitt: Probably worth relating these to the print callback?

```suggestion
type JsPrintCallback<'env> = Function<'env, FnArgs<(&'static str, String)>, ()>;
type JsPrintCallbackRef = FunctionRef<FnArgs<(&'static str, String)>, ()>;
```

davidhewitt (approved):
Thanks, this looks good to me! Just one tiny nit on a type alias name, which I think will help for the future...

PR #157, opened 2026-02-13
`re` module implementation

### LINKED ISSUE #132: `re` module impl in scope?
Is `re` part of the goal/scope? Because with it Monty could become THE sandbox layer for RLMs

Embers-of-the-Fire:
This PR introduces a subset of Python’s `re` module. It might close #132 as a partial implementation.

**Full API list**
- `re.Match`
  - `bool`, `type`, `__eq__`, `__repr__`
  - `.group()`, `.groups()`
  - `.span()`, `.end()`
  - `.string`
- `re.Pattern`
  - `.pattern`
  - `.flags`
  - Shares logic with module-level functions
- `re.IGNORECASE`, `re.MULTILINE`, `re.DOTALL`
- `re.compile()`, `re.match()`, `re.fullmatch()`, `re.findall()`, `re.sub()`
- `re.PatternError` ([notes](#repatternerror-behavior))

**Unresolved questions**
- [x] [`re.PatternError` behavior](#repatternerror-behavior)
- [x] [Regex behavior](#regex-behavior)

## `re.PatternError` behavior

The `monty` implementation introduces `re.PatternError` as a plain error with only a test.  
The `monty-python` binding falls back to `RuntimeError` because retrieving `re.PatternError` is awkward unless we do something like `py.import("re")?.get_attr("PatternError")?.call1((msg,))`.

## Regex behavior

By default, this PR uses the `regex` crate, which does not support backreferences, lookahead, etc.  
Switching to `fancy_regex` could more closely match CPython’s default behavior, but might introduce ReDoS risk.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,10 +36,11 @@ What Monty **can** do:
 - Control resource usage - Monty can track memory usage, allocations, stack depth, and execution time and cancel execution if it exceeds preset limits
 - Collect stdout and stderr and return it to the caller
 - Run async or sync code on the host via async or sync code on the host
+- Use a small subset of the standard library: `sys`, `os`, `typing`, `asyncio`, `re`, `datetime` (soon), `dataclasses` (soon), `json` (soon)
 
 What Monty **cannot** do:
 
-- Use the standard library (except a few select modules: `sys`, `typing`, `asyncio`, `dataclasses` (soon), `json` (soon))
+- Use the rest of the standard library
 - Use third party libraries (like Pydantic), support for external python library is not a goal
 - define classes (support should come soon)
 - use match statements (again, support should come soon)


Embers-of-the-Fire:
@samuelcolvin ReDoS stems from backtracking features, so just one more implementation won’t work. IIRC google's `re2` avoids those constructs entirely like `regex`. So it’s a trade‑off between a fast, stable engine and full compatibility.

crates/monty-python/src/exceptions.rs
```diff
@@ -408,6 +408,12 @@ pub fn exc_monty_to_py(py: Python<'_>, exc: MontyException) -> PyErr {
         ExcType::FileExistsError => exceptions::PyFileExistsError::new_err(msg),
         ExcType::IsADirectoryError => exceptions::PyIsADirectoryError::new_err(msg),
         ExcType::NotADirectoryError => exceptions::PyNotADirectoryError::new_err(msg),
+        // TODO: the error should be `re.PatternError` instead of RuntimeError,
+        //       but since the error class is not natively supported by PyO3,
+        //       we map it to RuntimeError for now.
+        //       To fetch for `re.PatternError`, we would need to run
+        //       py.import("re")?.getattr("PatternError")?.downcast()?,call1((msg,))?
```
samuelcolvin: copy the logic from `get_frozen_instance_error`.

crates/monty/src/types/re_pattern.rs
```diff
@@ -0,0 +1,506 @@
...
+    /// The original Python regex pattern string.
+    pattern: String,
+    /// Python regex flags bitmask (IGNORECASE=2, MULTILINE=8, DOTALL=16).
+    flags: u32,
+    /// The compiled Rust regex for `search` / `findall` / `sub` (unanchored).
+    compiled: Regex,
+    /// Compiled regex anchored at start for `match` operations.
+    compiled_match: Regex,
+    /// Compiled regex anchored at both ends for `fullmatch` operations.
+    compiled_fullmatch: Regex,
```
samuelcolvin: do we need three regexes? If not an enum might work better.

crates/monty/src/types/re_pattern.rs
```diff
@@ -0,0 +1,506 @@
...
+    ///
+    /// Follows CPython's semantics:
+    /// - No capture groups: returns a list of matched strings
+    /// - One capture group: returns a list of the group's matched strings
+    /// - Multiple capture groups: returns a list of tuples of matched strings
+    pub fn findall(&self, text: &str, heap: &mut Heap<impl ResourceTracker>) -> RunResult<Value> {
+        let cap_count = self.compiled.captures_len();
+        let mut results = Vec::new();
+
+        if cap_count <= 1 {
```
samuelcolvin: match statement might be very slightly more efficient?

samuelcolvin (commented):
Partial review since I have to go, but this looks great overall.

I would suggest we use `fancy-regex` with `backtrack_limit()` set to something reasonable, with that the dos risk seems acceptable.

Embers-of-the-Fire:
@samuelcolvin apologies for disturbing, but I wish to confirm the codecov. Certain sections of code prove difficult to construct during testing, so I am uncertain whether they require coverage.

crates/monty-python/src/re.rs
samuelcolvin: this doesn't need it's own file, put it in `exceptions.rs`.

crates/monty/src/modules/re.rs
```diff
@@ -0,0 +1,313 @@
+//! Implementation of the `re` module.
+//!
+//! Provides regular expression matching operations modeled after Python's `re` module.
+//! Uses the Rust `regex` crate, which guarantees linear-time matching (DFA-based),
```
samuelcolvin: make sure comments are up-to-date, we're now using `fancy-regex`, not just `regex`.

crates/monty/src/modules/re.rs
```diff
@@ -0,0 +1,313 @@
...
+        heap,
+        interns,
+    );
+    module.set_attr(
+        StaticStrings::MultilineFlag,
+        Value::Int(i64::from(MULTILINE)),
+        heap,
+        interns,
+    );
+    module.set_attr(StaticStrings::DotallFlag, Value::Int(i64::from(DOTALL)), heap, interns);
```
samuelcolvin: I don't see support `re.I` and the other aliases?

PR #159, opened 2026-02-13
Fix `TypeError` when comparing heap-allocated strings

sathish-t:
## Bug

A human user encountered this bug while using Monty via the JavaScript bindings. This PR was authored with Claude Code to diagnose the root cause and implement the fix.

`sorted()`, `min()`, `max()`, and comparison operators (`<`, `>`, `<=`, `>=`) fail with:

```
TypeError: '<' not supported between instances of 'str' and 'str'
```

when operating on strings created at runtime (e.g. from `str.split()`), even though the same operations work fine on string literals.

### Reproduction

```javascript
const { Monty, runMontyAsync } = require("@pydantic/monty");

// Works: sorted on literal strings
const m1 = new Monty('sorted(["banana", "apple", "cherry"])', { externalFunctions: [] });
const r1 = await runMontyAsync(m1, { limits: { maxDurationSecs: 5 } });
// => ["apple", "banana", "cherry"]

// Fails: sorted on strings from split()
const m2 = new Monty('sorted("banana,apple,cherry".split(","))', { externalFunctions: [] });
const r2 = await runMontyAsync(m2, { limits: { maxDurationSecs: 5 } });
// => TypeError: '<' not supported between instances of 'str' and 'str'
```

Equivalent Python reproduction:

```python
sorted(['banana', 'apple', 'cherry'])          # works (InternString literals)
sorted('banana,apple,cherry'.split(','))        # fails (heap-allocated Str)
```

## Root Cause

`Value::py_cmp()` in `value.rs` only handled string ordering for `InternString` vs `InternString` (compile-time string literals). String literals are interned at compile time, but strings created at runtime (from `split()`, concatenation, etc.) are heap-allocated as `Value::Ref(HeapData::Str)`.

The `(Ref, Ref)` match arm in `py_cmp()` only compared `LongInt` values, returning `Ok(None)` for any other heap type — which `sorted()` and `min()`/`max()` interpret as "comparison not supported", triggering the `TypeError`.

Note: `py_eq()` already handled all three string representation combinations correctly — `py_cmp()` was simply missing the same coverage.

## Fix

Extended `Value::py_cmp()` with three new comparison cases:

1. **`Ref(Str)` vs `Ref(Str)`** — two heap-allocated strings (added to the existing `Ref` vs `Ref` arm alongside `LongInt`)
2. **`InternString` vs `Ref(Str)`** — interned literal compared with heap string
3. **`Ref(Str)` vs `InternString`** — heap string compared with interned literal

## Tests

Added tests to existing test files:
- `builtin__more_iter_funcs.py` — `sorted()`, `min()`, `max()` with heap strings from `split()`
- `compare__mixed_types.py` — direct `<`/`>`/`<=`/`>=` between heap-allocated and interned strings

All 760 test cases pass, plus the full `make test-ref-count-panic` suite.

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/monty/src/value.rs
+++ b/crates/monty/src/value.rs
@@ -291,24 +291,32 @@ impl PyTrait for Value {
                     Ok(None)
                 }
             }
-            // LongInt vs LongInt comparison
+            // Ref vs Ref comparison: handles LongInt and Str
             (Self::Ref(id1), Self::Ref(id2)) => {
-                let is_longint1 = matches!(heap.get(*id1), HeapData::LongInt(_));
-                let is_longint2 = matches!(heap.get(*id2), HeapData::LongInt(_));
-                if is_longint1 && is_longint2 {
-                    Ok(heap.with_two(*id1, *id2, |_heap, left, right| {
-                        if let (HeapData::LongInt(a), HeapData::LongInt(b)) = (left, right) {
-                            a.inner().partial_cmp(b.inner())
-                        } else {
-                            None
-                        }
-                    }))
+                Ok(heap.with_two(*id1, *id2, |_heap, left, right| match (left, right) {
+                    (HeapData::LongInt(a), HeapData::LongInt(b)) => a.inner().partial_cmp(b.inner()),
+                    (HeapData::Str(a), HeapData::Str(b)) => a.as_str().partial_cmp(b.as_str()),
+                    _ => None,
+                }))
+            }
+            // Interned string comparisons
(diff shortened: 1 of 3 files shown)


samuelcolvin:
Thanks so much for this, and especially for this clear sign posting:

> This PR was authored with Claude Code to diagnose the root cause and implement the fix.

Also, sorry for the slow reply, I was on PTO last week.

PR #171, opened 2026-02-16
Implement the datetime plan with fixed offsets to start

mwildehahn:
## Scope
This PR is intentionally kept narrow to make review tractable.

## Deferred From This PR
These changes were removed from the diff against `main` and will land in a follow-up PR:
- `crates/monty-typeshed/vendor/typeshed/**`
- `crates/monty/src/bytecode/vm/async_exec.rs`
- `crates/monty/src/bytecode/vm/binary.rs`
- `crates/monty/src/bytecode/vm/call.rs`
- `crates/monty/src/bytecode/vm/compare.rs`
- `crates/monty/src/bytecode/vm/mod.rs`
- `crates/monty/src/bytecode/vm/scheduler.rs`
- `crates/monty/src/types/module.rs`

Dependent cleanups were also reverted to keep the branch compiling with the rollback:
- `crates/monty/src/repl.rs`
- `crates/monty/src/run_progress.rs`

## What Those Deferred VM/Module Changes Were Trying To Do
1. Preserve hidden resume metadata for pending `datetime.now` OS calls so async future resolution can reconstruct `date.today()`, naive `datetime.now()`, and fixed-offset aware `datetime.now(tz=...)` correctly.
2. Improve CPython parity for datetime/timedelta arithmetic and comparison edge cases (overflow paths and naive/aware error handling).
3. Change type/module method dispatch so module-exposed builtin types (notably `datetime.date` / `datetime.datetime`) handle class methods consistently.
4. Centralize some comparison and unary-op edge handling inside VM paths.

## Deferred Regression Coverage
The intended failing cases are captured as commented tests in:
- `crates/monty/test_cases/datetime__core.py`

Those snippets are intentionally commented out in this PR and should be re-enabled in the follow-up PR that restores the deferred VM/module work.

### DIFF
--- a/crates/monty/src/types/module.rs
+++ b/crates/monty/src/types/module.rs
@@ -161,7 +161,10 @@ impl Module {
                 func.drop_with_heap(args_guard.heap());
                 Err(ExcType::type_error("module attribute is not callable"))
             }
-            None => Err(ExcType::attribute_error_module(interns.get_str(self.name), attr.as_str(interns))),
+            None => Err(ExcType::attribute_error_module(
+                interns.get_str(self.name),
+                attr.as_str(interns),
+            )),
         }
     }
 }


crates/monty/src/bytecode/vm/call.rs
```diff
@@ -797,10 +798,16 @@ fn call_type_method(
     args: ArgValues,
     heap: &mut Heap<impl ResourceTracker>,
     interns: &Interns,
-) -> Result<Value, RunError> {
+) -> Result<AttrCallResult, RunError> {
     match (t, method_id) {
-        (Type::Dict, m) if m == StaticStrings::Fromkeys => return dict_fromkeys(args, heap, interns),
-        (Type::Bytes, m) if m == StaticStrings::Fromhex => return bytes_fromhex(args, heap, interns),
+        (Type::Dict, m) if m == StaticStrings::Fromkeys => {
```
samuelcolvin: we should probably move these methods into `dict.rs` etc. rather than have the logic here.

crates/monty/src/bytecode/vm/compare.rs
```diff
@@ -150,3 +164,16 @@ impl<T: ResourceTracker> VM<'_, '_, T> {
         }
     }
 }
+
+/// Returns datetime awareness (`true` for aware, `false` for naive) for datetime values.
+///
+/// Returns `None` when the value is not a datetime reference.
+fn datetime_awareness(value: &Value, heap: &crate::heap::Heap<impl ResourceTracker>) -> Option<bool> {
```
samuelcolvin: I think this method is duplicated.

crates/monty/src/types/date.rs
samuelcolvin: there's a lot of logic here. I know I recommended speedate, but would this be simpler if we used chrono?

I initially suggested not to use chrono because of it's leap second stuff, but maybe it would be cleaner than this???

crates/monty/src/types/date.rs
```diff
@@ -0,0 +1,408 @@
...
+    heap::{Heap, HeapData, HeapId},
+    intern::Interns,
+    os::OsFunction,
+    resource::{DepthGuard, ResourceError, ResourceTracker},
+    types::{AttrCallResult, PyTrait, TimeDelta, Type, timedelta},
+    value::Value,
+};
+
+/// `datetime.date` storage backed directly by `speedate`.
+pub(crate) type Date = speedate::Date;
```
samuelcolvin: you should wrapp the speedate Date (or chrono date) like

```
pub(crate) struct Date(speedate::Date);
```

Then you can implement (de)serialization directly here.

crates/monty-typeshed/vendor/typeshed/stdlib/builtins.pyi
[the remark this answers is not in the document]
mwildehahn: @samuelcolvin to get tests passing I had to regen typeshed. LMK if there is a better way to do this or if you still want me to remove these.

PR #173, opened 2026-02-17
better Pydantic AI example

samuelcolvin:


samuelcolvin:
@claude fix the formatting of the example that's causing CI to fail.

PR #178, opened 2026-02-17
when MontyRepl::start throws an exception, we want to also keep the Repl around

tkxue:
Right now, when MontyRepl:: start throws an exception, at https://github.com/pydantic/monty/blob/main/crates/monty/src/repl.rs#L326 , the repl is consumed due to the type signature.

```
    pub fn start(self, code: &str, print: &mut PrintWriter<'_>) -> Result<ReplProgress<T>, MontyException> {

```

This PR changes the signature to:

```
    pub fn start(self, code: &str, print: &mut PrintWriter<'_>) -> Result<ReplProgress<T>, (MontyRepl<T>, MontyException)> {

```

so we can keep the MontyRepl.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,7 @@ dbg_macro = "warn"
 use_self = "warn"
 allow_attributes = "warn"
 undocumented_unsafe_blocks = "warn"
+redundant_clone = "warn"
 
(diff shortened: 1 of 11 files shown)


tkxue:
I think #187 is better. I'm going to close mine in favor of #187

Issue #163, opened 2026-02-16
Common Python patterns LLMs use that don't work atm


adtyavrdhn:
**v0.0.5, Python 3.12**

- `sorted([3,1,2], key=lambda x: -x)` — keyword args not supported
- `sorted([3,1,2], reverse=True)` — same
- `sorted([(2,'b'),(1,'a')])` — tuple comparison fails inside sorted
- `{1,2} | {3,4}` / `&` / `-` — set operators not supported
- `d["a"]["b"] = 2` — chained subscript assignment fails

davidhewitt:
I believe these are all done.

PR #180, opened 2026-02-17
implement calling of user-defined functions

davidhewitt:
This enables `list.sort()` and `map()` builtins to call user-defined functions. This works by passing the `vm` through the relevant code paths, and adding a way to re-enter the `run()` loop when calling those functions.


samuelcolvin:
this would be a big win, we need to implement it for `sorted`, see #163.

PR #184, opened 2026-02-18
support sorted kwargs

samuelcolvin:
part of #163.

Plant from claude:

---

# Plan: Add `key=` and `reverse=` support to `sorted()`

## Context

`sorted()` currently rejects all keyword arguments (`kwargs.not_supported_yet()`). The goal is to support `sorted([3,1,2], key=lambda x: -x)` and `sorted([3,1,2], reverse=True)`, including user-defined key functions (lambdas).

The key architectural change: extend `BuiltinsFunctions::call` to take a VM context and return `CallOutcome` (renamed from `AttrCallResult`), so builtins can call user-defined functions via `VM::call_sync`. Business logic stays in the builtin files.

### 1. Rename `AttrCallResult` → `CallOutcome`

Use `/fastmod` to rename across the codebase. `CallOutcome` is more general since it's used by builtins, modules, heap types, and attribute calls — not just attribute calls.

**Files:** ~15 files across `types/`, `modules/`, `heap.rs`, `value.rs`, `exception_private.rs`, `bytecode/vm/call.rs`

### 2. Extend `BuiltinsFunctions::call` signature

**File:** `crates/monty/src/builtins/mod.rs`

```rust
impl BuiltinsFunctions {
    /// Primary call method with VM context for full function calling capability.
    pub(crate) fn call<T: ResourceTracker>(
        self,
        vm: &mut VM<'_, '_, T>,
        args: ArgValues,
    ) -> RunResult<CallOutcome> {
        match self {
            // Sorted needs VM for call_sync (key functions)
            Self::Sorted => sorted::builtin_sorted(vm, args),
            // All other builtins delegate to call_basic
            _ => self.call_basic(vm.heap, args, vm.interns, vm.print_writer)
                     .map(CallOutcome::Value),
        }
    }

    /// Basic call without VM context.
    /// Used by call_key_function in list.sort where VM isn't available.
    pub(crate) fn call_basic(
        self,
        heap: &mut Heap<impl ResourceTracker>,
        args: ArgValues,
        interns: &Interns,
        print_writer: &mut PrintWriter<'_>,
    ) -> RunResult<Value> {
        // Existing dispatch logic unchanged
        match self { ... }
    }
}
```

This keeps most builtin implementations unchanged. Only builtins needing VM access (sorted, later map) are handled in `call` directly.

Similarly update `Builtins::call` to take `&mut VM` and return `RunResult<CallOutcome>`, delegating `Type::call` through `.map(CallOutcome::Value)`.

### 3. Make VM fields accessible & implement `ContainsHeap`

**File:** `crates/monty/src/bytecode/vm/mod.rs`

Make `heap`, `interns`, `print_writer` fields `pub(crate)` so builtins receiving `&mut VM` can access them directly (enabling split borrows that avoid conflicts with `call_sync`).

Implement `ContainsHeap` for VM so `defer_drop!(value, vm)` works:
```rust
impl<T: ResourceTracker> ContainsHeap for VM<'_, '_, T> {
    type ResourceTracker = T;
    fn heap_mut(&mut self) -> &mut Heap<T> { self.heap }
}
```

### 4. Add `VM::call_sync` for calling functions synchronously

**File:** `crates/monty/src/bytecode/vm/mod.rs`

Add `subcall_depth: Option<usize>` field to VM. Add method:

```rust
pub(crate) fn call_sync(&mut self, callable: &Value, args: ArgValues) -> RunResult<Value> {
    let callable_clone = callable.clone_with_heap(self.heap);
    let result = self.call_function(callable_clone, args)?;
    match result {
        CallResult::Push(value) => Ok(value),
        CallResult::FramePushed => {
            let prev = self.subcall_depth;
            self.subcall_depth = Some(self.frames.len() - 1);
            let result = self.run();
            self.subcall_depth = prev;
            match result {
                Ok(FrameExit::Return(value)) => Ok(value),
                Err(e) => Err(e),
                _ => Err(RunError::internal("unexpected exit in call_sync")),
            }
        }
        _ => Err(ExcType::type_error("key function cannot be an external function")),
    }
}
```

### 5. Modify `ReturnValue` handler for subcall boundary

**File:** `crates/monty/src/bytecode/vm/mod.rs`

After `self.pop_frame()` in the `ReturnValue` opcode handler, check:
```rust
if Some(self.frames.len()) == self.subcall_depth {
    return Ok(FrameExit::Return(value));
}
```

### 6. Modify `handle_exception` for subcall boundary

**File:** `crates/monty/src/bytecode/vm/exceptions.rs`

Before unwinding past the subcall boundary frame, stop and return the error:
```rust
if let Some(depth) = this.subcall_depth {
    if this.frames.len() <= depth + 1 {
        this.pop_frame();
        return Some(error);
    }
}
```

### 7. Update VM call sites

**File:** `crates/monty/src/bytecode/vm/call.rs`

- `call_function`: Change `Value::Builtin(builtin) => builtin.call(self, args)` (pass VM, handle `CallOutcome`)
- `exec_call_builtin_function`: Change to use `builtin.call(self, args)`, return type becomes `Result<CallResult, RunError>`
- Update `CallBuiltinFunction` opcode dispatch in `mod.rs` to use `handle_call_result!` instead of direct push

Remove `CallResult` enum — replace with `CallOutcome` plus `FramePushed` (either merge or keep `CallResult` as a simple conversion from `CallOutcome` plus `FramePushed`). Actually, keep `CallResult` as VM-internal with `FramePushed` variant, and convert from `CallOutcome` via the existing `From` impl (renamed).

### 8. Implement `builtin_sorted` with kwargs

**File:** `crates/monty/src/builtins/sorted.rs`

```rust
pub fn builtin_sorted<T: ResourceTracker>(
    vm: &mut VM<'_, '_, T>,
    args: ArgValues,
) -> RunResult<CallOutcome> {
    let (positional, kwargs) = args.into_parts();
    // Parse kwargs: key and reverse (reuse extract pattern from list.sort)
    // Collect items from iterable
    // If key_fn provided and not None:
    //   for each item: vm.call_sync(&key_fn, One(item.clone_with_heap(vm.heap)))
    // Sort using index-based permutation (same pattern as do_list_sort)
    // with reverse flag
    // Return sorted list as CallOutcome::Value(...)
}
```

Key implementation details:
- Use `defer_drop!` with `vm` (works because VM implements `ContainsHeap`)
- Sequential borrows: `vm.heap` for cloning items, then `vm.call_sync()` for key functions — borrows don't overlap
- Sorting uses index permutation pattern from `do_list_sort` (`crates/monty/src/types/list.rs:800-895`)
- `key=None` treated as no key function (same as Python)
- Handle `DepthGuard` for `py_cmp` calls during sorting

### 9. Update `call_key_function` in list.rs

**File:** `crates/monty/src/types/list.rs`

Change from calling `builtin.call(heap, args, interns, print_writer)` to `builtin.call_basic(heap, args, interns, print_writer)`. This preserves the current behavior (builtin-only key functions for `list.sort`).

### 10. Add tests

**File:** `crates/monty/test_cases/builtin__sorted_kwargs.py` (new, separate file because kwargs path uses different compilation)

```python
# === reverse=True ===
assert sorted([3, 1, 2], reverse=True) == [3, 2, 1], 'reverse int list'
assert sorted([1, 2, 3], reverse=False) == [1, 2, 3], 'reverse=False is default'
assert sorted([], reverse=True) == [], 'reverse empty'
assert sorted('cab', reverse=True) == ['c', 'b', 'a'], 'reverse string iterable'

# === key function (lambda) ===
assert sorted([3, 1, 2], key=lambda x: -x) == [3, 2, 1], 'key negate'
assert sorted(['banana', 'apple', 'cherry'], key=lambda s: len(s)) == ['apple', 'banana', 'cherry'], 'key len'
assert sorted([(1, 'b'), (2, 'a'), (1, 'a')], key=lambda t: t[1]) == [(2, 'a'), (1, 'a'), (1, 'b')], 'key element'

# === key function (builtin) ===
assert sorted([3, -1, 2], key=abs) == [-1, 2, 3], 'key abs'

# === key + reverse ===
assert sorted([3, 1, 2], key=lambda x: -x, reverse=True) == [1, 2, 3], 'key + reverse'

# === key=None (default) ===
assert sorted([3, 1, 2], key=None) == [1, 2, 3], 'key None default'
assert sorted([3, 1, 2], key=None, reverse=True) == [3, 2, 1], 'key None + reverse'
```

## Files summary

| File | Change |
|------|--------|
| ~15 files | Rename `AttrCallResult` → `CallOutcome` |
| `crates/monty/src/builtins/mod.rs` | Split `call` into `call` (VM) + `call_basic` (old) |
| `crates/monty/src/builtins/sorted.rs` | Full rewrite with VM, kwargs, key/reverse support |
| `crates/monty/src/bytecode/vm/mod.rs` | `pub(crate)` fields, `ContainsHeap` impl, `subcall_depth`, `call_sync`, `ReturnValue` check |
| `crates/monty/src/bytecode/vm/exceptions.rs` | Subcall boundary in `handle_exception` |
| `crates/monty/src/bytecode/vm/call.rs` | Update `call_function`, `exec_call_builtin_function` for new signature |
| `crates/monty/src/types/list.rs` | `call_key_function` uses `call_basic` |
| `crates/monty/test_cases/builtin__sorted_kwargs.py` | New test file |

## Verification

```bash
# Run new sorted kwargs tests
cargo test -p monty --test datatest_runner --features ref-count-panic sorted_kwargs

# Run existing sorted tests (regression check)
cargo test -p monty --test datatest_runner --features ref-count-panic builtin__more_iter_funcs

# Run ref-count-panic tests (leak detection)
make test-ref-count-panic

# Format and lint
make format-rs && make lint-rs
```

### DIFF
--- a/crates/monty/src/builtins/map.rs
+++ b/crates/monty/src/builtins/map.rs
@@ -76,7 +76,7 @@ pub fn builtin_map(
         [] => {
             while let Some(item) = first_iter.for_next(heap, interns)? {
                 let args = ArgValues::One(item);
-                out.push(builtin.call(heap, args, interns, print_writer)?);
+                out.push(builtin.call_basic(heap, args, interns, print_writer)?);
             }
         }
         // map(f, iter1, iter2)
@@ -87,7 +87,7 @@ pub fn builtin_map(
                     break;
                 };
                 let args = ArgValues::Two(arg1, arg2);
-                out.push(builtin.call(heap, args, interns, print_writer)?);
+                out.push(builtin.call_basic(heap, args, interns, print_writer)?);
             }
         }
         // map(f, iter1, iter2, *iterables)
@@ -108,7 +108,7 @@ pub fn builtin_map(
                 kwargs: KwargsValues::Empty,
             };
 
-            out.push(builtin.call(heap, args, interns, print_writer)?);
+            out.push(builtin.call_basic(heap, args, interns, print_writer)?);
         },
     }
 
--- a/crates/monty/src/builtins/mod.rs
+++ b/crates/monty/src/builtins/mod.rs
@@ -36,12 +36,13 @@ use strum::{Display, EnumString, FromRepr, IntoStaticStr};
 
 use crate::{
     args::ArgValues,
+    bytecode::VM,
     exception_private::{ExcType, RunResult},
     heap::Heap,
     intern::Interns,
     io::PrintWriter,
     resource::ResourceTracker,
-    types::Type,
+    types::{CallOutcome, Type},
     value::Value,
 };
 
@@ -60,22 +61,34 @@ pub(crate) enum Builtins {
 }
 
 impl Builtins {
-    /// Calls this builtin with the given arguments.
+    /// Calls this builtin with full VM context, returning a `CallOutcome`.
     ///
-    /// # Arguments
-    /// * `heap` - The heap for allocating objects
-    /// * `args` - The arguments to pass to the callable
-    /// * `interns` - String storage for looking up interned names in error messages
-    /// * `print` - The print for print output
-    pub fn call(
+    /// Builtin functions that need VM access (e.g., `sorted()` with key functions)
+    /// are dispatched through `BuiltinsFunctions::call`. All others delegate to
+    /// `call_basic` which only needs heap/interns/print_writer.
+    ///
+    /// Exception types and type constructors always complete synchronously.
+    pub fn call<T: ResourceTracker>(self, vm: &mut VM<'_, '_, T>, args: ArgValues) -> RunResult<CallOutcome> {
+        match self {
+            Self::Function(b) => b.call(vm, args),
+            Self::ExcType(exc) => exc.call(vm.heap, args, vm.interns).map(CallOutcome::Value),
+            Self::Type(t) => t.call(vm.heap, args, vm.interns).map(CallOutcome::Value),
+        }
+    }
+
+    /// Calls this builtin without VM context.
+    ///
+    /// Used by `map()` and other contexts where builtins are called as function
+    /// arguments without needing VM access.
+    pub fn call_basic(
         self,
         heap: &mut Heap<impl ResourceTracker>,
         args: ArgValues,
         interns: &Interns,
-        print: &mut PrintWriter<'_>,
+        print_writer: &mut PrintWriter<'_>,
     ) -> RunResult<Value> {
         match self {
-            Self::Function(b) => b.call(heap, args, interns, print),
+            Self::Function(b) => b.call_basic(heap, args, interns, print_writer),
             Self::ExcType(exc) => exc.call(heap, args, interns),
             Self::Type(t) => t.call(heap, args, interns),
         }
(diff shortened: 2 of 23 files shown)


davidhewitt:
I implemented almost exactly the same thing in #180 - main differences in my PR:
- `evaluate_function` as a new function in VM which handles all callables both built-in and udf so that each caller doesn't need to care about the differences, plus raises a common error if external function attempts to suspend for now
- I store the recursion depth as a bool flag on the frame to stop at, not on the VM itself. I think if ever supporting suspecting inside these calls that will be necessary

PR #201, opened 2026-02-23
Fix os.environ panic in JS bindings

sathish-t:
## Summary

A human discovered that accessing `os.environ` via the JS bindings' `start()` method caused a **Rust panic that crashes the entire Node.js process**, instead of returning a graceful error. This PR fixes the panic and two related issues in how OS calls are handled in the JS bindings.

**Bug discovered by a human; diagnosed and fixed with Claude Code.**

## The Bug

When Python code accesses `os.environ` (or calls `os.getenv`, or any OS function), the Monty VM yields a `RunProgress::OsCall` back to the host, asking the host to handle the operation. The JS bindings had no handler for this and panicked:

```
thread '<unnamed>' panicked at crates/monty-js/src/monty_cls.rs:1010:13:
OS calls are not yet supported in the JS bindings: GetEnviron
fatal runtime error: failed to initiate panic, error 5, aborting
```

## Three Code Paths, Three Behaviors

There are three ways OS calls can be triggered in the JS bindings. Before this fix, they each behaved differently:

### Path 1: `run()` without external functions — already correct
This calls `MontyRun::run()` in the core Rust layer, which handles the entire execution loop internally. When it encounters an `OsCall` with no host callback, it feeds a `NotImplementedError` exception **back into the VM as a Python exception**. The VM raises it, and the core returns a `MontyException`. The JS napi layer wraps it as `JsMontyException`, and the TS wrapper throws a `MontyRuntimeError`. No changes needed here.

### Path 2: `run()` WITH external functions (`run_with_external_functions`) — fixed
This uses `start()` + loop internally. When it hit an `OsCall`, it returned `Err(Error::from_reason(...))` — a **plain napi `Error`**, not a `MontyRuntimeError`. Users couldn't catch this with `instanceof MontyRuntimeError` like they can for every other Monty error.

**Fix:** Now returns `Ok(Either::B(JsMontyException::new(...)))` with `NotImplementedError`, so the wrapper correctly throws a `MontyRuntimeError`.

### Path 3: `start()` — fixed (the panic)
This called `progress_to_result()` which had `panic!()` for `RunProgress::OsCall`. This crashes the Node.js process with no way to catch or recover.

**Fix:** Now returns `Either3::C(JsMontyException::new(...))` with `NotImplementedError`, which the wrapper's `wrapStartResult` converts to a thrown `MontyRuntimeError`.

### End result
All three paths now produce the same user-facing behavior: a `MontyRuntimeError` with `NotImplementedError` inside, matching how the Python bindings handle the same scenario.

## Changes

- **`crates/monty-js/src/monty_cls.rs`**: Replace `panic!()` in `progress_to_result()` with graceful `JsMontyException` returns for both `OsCall` and `ResolveFutures`. Fix `run_with_external_functions` `OsCall` arm to return `JsMontyException` instead of plain napi `Error`.
- **`crates/monty-js/__test__/exceptions.spec.ts`**: 3 new tests for `os.environ`/`os.getenv` via `run()` (with and without external functions).
- **`crates/monty-js/__test__/start.spec.ts`**: 2 new tests for `os.environ`/`os.getenv` via `start()`.

## Test plan

- [x] Reproduced the panic on JS side before fix (`start()` crashes Node.js process)
- [x] Verified Python side already handles this gracefully (returns `NotImplementedError`)
- [x] All 294 JS tests pass (including 5 new tests)
- [x] All Rust tests pass (`make test-ref-count-panic`: 0 failures)
- [x] `make lint-rs` clean, `make lint-js` clean


davidhewitt (approved):
Thanks!

Issue #210, opened 2026-02-27
iterable into dict


samuelcolvin:
We need to support

```py
model_info = dict(zip(headers, row_data))
```

Which currently returns 

```
pydantic_monty.MontyRuntimeError: TypeError: 'list' object is not iterable
```

PR #212, opened 2026-02-28
feat(dict): support dict(iterable, **kwargs) and fix dict(zip(...))

FarhanAliRaza:
- extend dict() constructor to accept iterable key/value pairs and keyword args
- refactor dict() and dict.update() to share merge logic
- preserve existing dict copy behavior and pair-length error semantics
- add type__ops coverage for dict(zip(...)), zip truncation, kwargs-only, and kwargs precedence

Fixed https://github.com/pydantic/monty/issues/210


davidhewitt (approved):
Thanks!

PR #235, opened 2026-03-03
Repl improvements

samuelcolvin:


crates/monty-python/src/serialization.rs
```diff
@@ -0,0 +1,761 @@
...
+        )));
+    }
+
+    let stored_hash = &bytes[2..HEADER_SIZE];
+    let payload = &bytes[HEADER_SIZE..];
+
+    let computed_hash = Sha256::digest(payload);
+    if computed_hash.as_slice() != stored_hash {
+        return Err(PyValueError::new_err("Serialized data integrity check failed"));
+    }
```
davidhewitt: What's the purpose of the integrity check? I would assume that `postcard` would just catch bugs anyway.

PR #221, opened 2026-03-03
feat(collections): add Counter support

tristanang:
Summary:
- add a new built-in collections module
- implement collections.Counter([source], **kwargs) with a performant dict-backed first pass
- wire module/function dispatch for collections
- align Counter.__init__ too-many-positional-args error message
- add consolidated datatest coverage in crates/monty/test_cases/import__collections_counter.py

Validation:
- make format-rs
- PYO3_PYTHON=/Users/tristanang/.local/bin/python3 make lint-rs
- make lint-py
- make test-cases

PR #250, opened 2026-03-08
feat: expose collections.namedtuple via a callable factory

tristanang:
## Summary
- add a builtin `collections` module and expose `namedtuple`
- implement a heap-backed `NamedTupleFactory` that returns the existing `NamedTuple` runtime value
- align default namedtuple factory repr behavior with CPython for omitted and `module=None` cases
- add interpreter and REPL coverage for imports, construction, defaults, rename behavior, and error cases

## Notes
- the returned object is a callable factory with class-like repr behavior rather than a true Python class
- instances still use the existing `NamedTuple` runtime representation
- I did not add a benchmark for namedtuple in this change

## Verification
- `PYO3_PYTHON=$(which python3) make format-rs`
- `PYO3_PYTHON=$(which python3) make lint-rs`
- `PYO3_PYTHON=$(which python3) cargo check -p monty`
- `PYO3_PYTHON=$(which python3) cargo test -p monty --test repl -- --nocapture`
- `uv run sh -lc 'PYO3_PYTHON="$VIRTUAL_ENV/bin/python3" cargo test -p monty --test datatest_runner namedtuple__collections -- --nocapture'`

## Environment Note
- in this worktree, running PyO3-backed test binaries via the project `.venv` avoids the interpreter mismatch that caused the earlier embedded-CPython `encodings` failure


davidhewitt:
Similar to #221, will close for now. Thanks again for the PR.

PR #251, opened 2026-03-08
fix(vm): fix partial future resolution panics in mixed gathers

runyaga:
## Summary

Fixes two panics in `async_exec.rs` that occur during incremental resolution of `asyncio.gather()` when a gather mixes coroutine tasks with direct external calls.

Closes #240

## Bug 1: `prepare_current_task_after_resolve` — "no active frame" panic

After a partial resume where `load_ready_task_if_needed` saves the current task's context (draining frames), the next call to `prepare_current_task_after_resolve` still considers the task ready and attempts to push a value onto an empty frame stack. `vm.run()` then panics with "no active frame".

**Fix:** Early return `false` when `self.frames.is_empty()`, deferring to `load_ready_task_if_needed` to restore the task context.

## Bug 2: `resolve_future` gather path — premature gather completion

`std::mem::take(&mut gather.task_ids)` drains the task ID vec before checking whether all tasks have completed. If the gather isn't fully resolved yet, `handle_task_completion` later reads the empty vec, considers the gather vacuously complete (zero tasks = all done), and panics on unfilled result slots.

**Fix:** Clone `task_ids` instead of taking ownership, preserving the gather's internal state for subsequent completion checks.

## Test plan

Two new tests reproduce the exact panic conditions:

- `gather_mixed_coroutine_and_direct_external_partial_resolve` — mixed coroutine + direct external call with partial resolution (Bug 1)
- `gather_three_tasks_with_direct_external_memtake_corruption` — three-way gather exposing the `mem::take` corruption (Bug 2)

```
test result: ok. 21 passed; 0 failed (asyncio)
test result: ok. 818 passed; 0 failed (datatest_runner, ref-count-panic)
```

All existing tests pass across `ref-count-panic`, `ref-count-return`, and no-features configurations.


crates/monty/src/bytecode/vm/async_exec.rs
```diff
@@ -778,18 +778,14 @@ impl<T: ResourceTracker> VM<'_, '_, T> {
...
                     gather.pending_calls.retain(|&cid| cid != call_id);
-                    // Take task_ids to avoid clone - we're checking completion so gather may be destroyed
-                    (
-                        gather.pending_calls.is_empty(),
-                        std::mem::take(&mut gather.task_ids),
-                        gather.waiter,
-                    )
+                    // Clone task_ids — mem::take would leave an empty vec that makes
+                    // handle_task_completion falsely consider the gather vacuously complete.
+                    (gather.pending_calls.is_empty(), gather.task_ids.clone(), gather.waiter)
```
davidhewitt: It looks like `gather.task_ids.clone()` is necessary only to work around borrow checker errors. It is probably possible to avoid cloning by instead doing something like `let scheduler = self.scheduler.as_ref().unwrap()` to get a borrow on the scheduler here, and moving the `all_tasks_complete` logic up here.

davidhewitt (commented):
Thanks for the PR, I think maybe opens some bigger questions.

PR #263, opened 2026-03-10
Fix forward function references at module scope

### LINKED ISSUE #183: Forward function references at module scope cause NameError
Really enjoying Monty, it's been a missing piece for me in a lot of applications. Great project.

Functions defined at module scope that reference other functions defined later in the same source get `NameError` at runtime, even though both functions exist by the time the call happens.

In CPython, names inside a function body are resolved at call time, not at definition time, so definition order doesn't matter. Monty appears to resolve them at definition/compile time.

## Minimal repro

```python
# test_cpython.py — works fine in CPython 3.12+
def foo():
    return bar()

def bar():
    return "hello from bar"

result = foo()
print(result)  # "hello from bar"
```

```python
# test_monty.py — fails in Monty v0.0.6
import pydantic_monty

code = """
def foo():
    return bar()

def bar():
    return "hello from bar"

result = foo()
result
"""

m = pydantic_monty.Monty(code)
print(m.run())
```

## Expected

Both scripts print `hello from bar`.

## Actual

CPython:

```
hello from bar
```

Monty:
```
NameError: name 'bar' is not defined
```

## Workaround

Define callees before callers (i.e. put `bar` above `foo`).

## Version

`pydantic-monty==0.0.6`, macOS ARM64, Python 3.14.

wsenn:
Closes #183

## Summary

- Pre-scan all module-level names before the main prepare pass so function bodies can resolve forward references
- In CPython, names inside function bodies are resolved at call time (via dictionary lookup), so `def foo(): return bar()` works even when `bar` is defined after `foo`
- Monty's prepare phase was snapshotting the global `name_map` when processing each function def, but the snapshot only contained names processed so far — causing forward references to fail with `NameError`

## Approach

Added a `pre_scan_module_names` method that reuses the existing `collect_scope_info_from_node` helper to collect all names that will be assigned at module level, then pre-allocates namespace slots for each. This runs before `prepare_nodes`, ensuring the `global_name_map` snapshot passed to each function body is complete.

By reusing `collect_scope_info_from_node`, the implementation:
- Avoids duplicating the node traversal logic (~25 lines vs ~100 in the initial version)
- Correctly handles walrus expressions, unpack targets, and control flow recursion
- Correctly skips `SubscriptAssign`/`SubscriptOpAssign` (which modify containers, not create new names)
- Stays in sync automatically when new node types are added

## Test plan

- [x] Added 5 test cases to `function__ops.py`: basic forward ref, chained forward refs, forward ref with arguments, forward ref to global variable, mutual recursion
- [x] All 887 existing test cases pass (1 pre-existing failure in `math__module.py` unrelated to this change)
- [x] `ref-count-panic` tests pass with no leaks
- [x] `cargo clippy` clean

### DIFF
--- a/crates/monty/src/prepare.rs
+++ b/crates/monty/src/prepare.rs
@@ -1,6 +1,7 @@
 use std::collections::hash_map::Entry;
 
 use ahash::{AHashMap, AHashSet};
+use indexmap::IndexSet;
 
 use crate::{
     args::{ArgExprs, CallArg, CallKwarg},
@@ -37,6 +38,22 @@ pub struct PrepareResult {
     pub nodes: Vec<PreparedNode>,
     /// The string interner containing all interned identifiers and filenames.
     pub interner: InternerBuilder,
+}
+
+trait AssignedNameCollector {
+    fn insert_name(&mut self, name: String);
+}
+
+impl AssignedNameCollector for AHashSet<String> {
+    fn insert_name(&mut self, name: String) {
+        self.insert(name);
+    }
+}
+
+impl AssignedNameCollector for IndexSet<String> {
+    fn insert_name(&mut self, name: String) {
+        self.insert(name);
+    }
 }
 
 /// Prepares parsed nodes for compilation by resolving names and building the initial namespace.
@@ -317,7 +334,7 @@ impl<'i> Prepare<'i> {
         // Collect all names that will be assigned at module level.
         // We reuse collect_scope_info_from_node which already handles all node types
         // including walrus expressions, unpack targets, and control flow recursion.
-        let mut assigned_names = AHashSet::new();
+        let mut assigned_names = IndexSet::new();
         let mut global_names = AHashSet::new();
         let mut nonlocal_names = AHashSet::new();
         for node in nodes {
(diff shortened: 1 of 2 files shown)


rsr5:
I created this issue (#183) and I've reviewed the approach. It looks clean and correct. The pre-scan reuses existing `collect_scope_info_from_node` infrastructure, keeps the change minimal, and the tests cover the important cases (chained refs, mutual recursion, etc.). Would be great to get this merged.

@davidhewitt would you be able to take a look please?

rsr5:
Thanks for picking this up @wsenn!

davidhewitt:
Thanks for the PR @wsenn and the ping @rsr5.

I have taken a look and I think this solves part of the problem but not the full picture when interacting with builtins. e.g. cpython handles the following code fine, but neither monty with this patch nor without works properly.

```python
def call_sum():
    # this will resolve as the builtin sum until the `def sum` later on
    return sum([1, 2, 3])

# with this patch, monty fails on this assert with `NameError`
assert call_sum() == 6


def sum(*args):
    return 42

# previously, monty would fail on this assert because it would still call the builtin `sum`
assert call_sum() == 42
```

davidhewitt:
I will explore possible solutions and either send review feedback to this PR, push to this PR, or push a new PR, whichever I think is easiest.

davidhewitt:
I opened #469 to address this case with a more comprehensive set of adjustment. Thanks again both!

PR #295, opened 2026-03-27
use dynamic dispatch for `ResourceTracker`

davidhewitt:
Fully vibe coded by claude, might be rubbish, I'm just curious what the perf looks like if we remove the infectious generic parameter.


davidhewitt:
Not convinced this is worth it for now.

PR #297, opened 2026-03-28
fix(repl): Add local/global store operation in comprehension expressions

### LINKED ISSUE #400: REPL List comprehension corrupting global func table?
one-shot works. repl fails

## ONE-SHOT
```
from pydantic_monty import Monty

# Simple lambda to avoid argument mismatch
host_func = lambda: "success"

# The verbatim sequence in a single one-shot script
code = '''
def x(): pass
[n for n in [5]]
print(f"Calling func...")
print(func())
'''

print("--- pydantic-monty One-Shot Reproduction ---")

try:
    m = Monty(code)
    # Run it once with external functions
    res = m.run(external_functions={"func": host_func})
    print(f"Success: {res}")
except Exception as e:
    print(f"Result: {e}")
```

## REPL
```
from pydantic_monty import MontyRepl

repl = MontyRepl()
ext = {"func": lambda: "ok"}

# Step 1: Initialize global names
repl.feed_run('def x(): pass', external_functions=ext)

# Step 2: Leak an integer into the name table
# In Monty, list comprehension variables leak to global scope
repl.feed_run('[n for n in [5]]', external_functions=ext)

print("--- Small Reproduction ---")
try:
    # Step 3: Call host function. 
    # It will fail because 'n' (5) collided with 'func'
    res = repl.feed_run('func()', external_functions=ext)
    print(f"Success: {res}")
except Exception as e:
    print(f"Reproduced: {e}")
    # Inspect what 'func' is now
    check = repl.feed_run('print(f"func is now: {type(func).__name__} with value {repr(func)}")')
    if check: print(check)
```

Embers-of-the-Fire:
Fixes #281 by force-clearing global slots left by list/generator comprehensions.

Fixes #400


davidhewitt:
Thanks again for this and sorry for the slow review. After a bunch of thinking I pushed an alternative implementation on this branch where the comprehension locals go directly in the VM stack, which seems to work well and leads to the most efficient bytecode.

PR #306, opened 2026-03-30
Implement strict mode for zip() function to enforce equal lengths of iterables

amitrechavia:
strict mode for zip()


davidhewitt:
Thanks, I pushed a copy of this with an additional commit for error handling in #324

PR #307, opened 2026-03-31
add context manager support

samuelcolvin:
Implements the Python context manager protocol (`__enter__`/`__exit__`) and the `open()` builtin function for sandboxed file I/O via OsCall.

The `with` statement compiles to bytecode using existing opcodes (Dup, CallAttr, exception table entries) with FinallyTarget integration for return/break/continue. The `open()` builtin yields OsCall(FileOpen) to the host, which returns file content as MontyObject::FileData. A new FileObject heap type provides read(), readline(), write(), close(), and the context manager methods.

Key changes:
- Parser: parse `with` statements, desugar multiple items to nested nodes
- Compiler: compile_with() emits setup/body/normal-exit/exception-handler bytecode
- Builtins: BuiltinsFunctions::call() returns CallResult (was Value) for OsCall support
- New types: FileObject, OsFunction::FileOpen/FileClose, MontyObject::FileData
- Test infra: virtual open() for CPython test compatibility


samuelcolvin:
replaced by #462

PR #308, opened 2026-04-01
Fix PGO and cross-arch CI builds failing on missing `typing_extensions`

penberg:
The `--no-deps` flag skips installing runtime dependencies, so `import pydantic_monty` fails with `ModuleNotFoundError`. Install `typing_extensions` explicitly before the `--no-deps` wheel install in both the PGO profiling step and the cross-architecture smoke test.


samuelcolvin:
Thanks so much!

PR #309, opened 2026-04-02
del statement support

kmad:
## `del` statement support

  Implements `del` across the full pipeline — parsing, prepare, bytecode compilation, and VM
  execution.

  ### What works

  - **`del x`** — unbinds variables (locals, globals, cells), subsequent access raises `NameError` or
   `UnboundLocalError` exactly as CPython does
  - **`del d['key']`** — dict item removal via `__delitem__`
  - **`del lst[i]`** — list item removal with negative index support
  - **`del a, b`** — multi-target, executed left-to-right
  - **Closures** — `del` on captured variables correctly propagates: outer scope gets
  `UnboundLocalError`, inner closures get `NameError` for free variables
  - **`nonlocal` + `del`** — deleting via `nonlocal` unbinds the cell in the enclosing scope

  ### New opcodes

  - `DeleteSubscr` — pops index + container, calls `__delitem__`
  - `DeleteLocalW` — wide variant of `DeleteLocal` for slots > 255
  - `DeleteCell` — stores `Undefined` in a closure cell

  All three appended to the enum to preserve discriminant stability.

  ### Bug fix (pre-existing)

  `load_global` was raising `UnboundLocalError` for undefined module-level variables that had
  previously been assigned. CPython always raises `NameError` for globals. Fixed.

  ### Not yet supported

  - `del obj.attr` — blocked on user-defined classes
 
Written in collaboration with claude-opus-4-6 and gpt-5.4-high

### DIFF
--- a/crates/monty/src/bytecode/compiler.rs
+++ b/crates/monty/src/bytecode/compiler.rs
@@ -3113,12 +3113,7 @@ impl<'a> Compiler<'a> {
                     if matches!(target.scope, NameScope::Local) {
                         self.code.register_assigned_local(slot);
                     }
-                    if let Ok(s) = u8::try_from(slot) {
-                        self.code.emit_u8(Opcode::DeleteLocal, s);
-                    } else {
-                        // Wide variant not implemented yet
-                        todo!("DeleteLocalW for slot > 255");
-                    }
+                    self.code.emit_delete_local(slot);
                 }
             }
             NameScope::Global => {
@@ -3126,10 +3121,8 @@ impl<'a> Compiler<'a> {
                 self.code.emit_u16(Opcode::DeleteGlobal, slot);
             }
             NameScope::Cell => {
-                // Delete cell not commonly needed
-                // For now, just store None
-                self.code.emit(Opcode::LoadNone);
-                self.compile_store(target);
+                self.code.register_local_name(slot, target.name_id);
+                self.code.emit_u16(Opcode::DeleteCell, slot);
             }
         }
     }
--- a/crates/monty/src/bytecode/compiler.rs
+++ b/crates/monty/src/bytecode/compiler.rs
@@ -3103,18 +3103,26 @@ impl<'a> Compiler<'a> {
     /// because module-level locals live in the globals array.
     fn compile_delete(&mut self, target: &Identifier) {
         let slot = u16::try_from(target.namespace_id().index()).expect("local slot exceeds u16");
+        self.code.set_location(target.position, None);
         match target.scope {
             NameScope::Local | NameScope::LocalUnassigned => {
+                self.code.register_local_name(slot, target.name_id);
                 if self.is_module_scope {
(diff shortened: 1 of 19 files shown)


kmad:
@samuelcolvin the performance regression has been resolved.

crates/monty/src/bytecode/vm/mod.rs
```diff
@@ -583,6 +644,17 @@ pub struct VM<'h, 'a, T: ResourceTracker> {
...
+    /// Global slots that have been explicitly deleted via `del`.
+    ///
+    /// Used by `load_global` to distinguish "deleted by `del`" (→ `NameError`) from
+    /// "unbound comprehension variable" (→ `UnboundLocalError`). Both appear as
+    /// `Value::Undefined` with `is_assigned_local` set, but CPython raises different
+    /// exception types for each case.
+    ///
+    /// Boxed and wrapped in `Option` to keep the VM struct small (8 bytes vs 48)
+    /// since `del` on globals is rare and the VM is accessed on every instruction.
+    deleted_globals: Option<Box<HashSet<u16>>>,
```
samuelcolvin, line 657: I don't really understand the point of `deleted_globals`. suspect we can remove it.

Even if there are subtle differences between cpython and monty on this, it seems like a price worth paying.

crates/monty/src/bytecode/vm/mod.rs
```diff
@@ -1951,3 +2127,63 @@ impl<T: ResourceTracker> ContainsHeap for VM<'_, '_, T> {
         self.heap
     }
 }
+
+#[cfg(test)]
```
samuelcolvin, line 2131: I'd rather not add tests in code unless absolutely necessary.

crates/monty/src/bytecode/compiler.rs
```diff
@@ -3088,25 +3103,26 @@ impl<'a> Compiler<'a> {
...
+                self.code.register_local_name(slot, target.name_id);
                 if self.is_module_scope {
                     self.code.emit_u16(Opcode::DeleteGlobal, slot);
-                } else if let Ok(s) = u8::try_from(slot) {
-                    self.code.emit_u8(Opcode::DeleteLocal, s);
                 } else {
-                    // Wide variant not implemented yet
-                    todo!("DeleteLocalW for slot > 255");
+                    if matches!(target.scope, NameScope::Local) {
+                        self.code.register_assigned_local(slot);
```
samuelcolvin, line 3114: I'm not clear why this is necessary?

crates/monty/src/bytecode/vm/mod.rs
samuelcolvin: If we really need this to be the same, maybe we could add a new variant to `Value`, e.g. `Value::GlobalDeleted` instead of adding this `deleted_globals`?

samuelcolvin (commented):
otherwise looking good.

samuelcolvin:
no reply.

PR #333, opened 2026-04-14
fix: support datetime.now() in standard execution (#330)

### LINKED ISSUE #330: `datetime.now()` not supported in standard execution
Monty cannot execute scripts that call `datetime.now()` in standard (non-suspendable) execution mode.

**Reproduce:**

```
$ cargo run -- -c "from datetime import datetime; print(datetime.now())"
error: NotImplementedError: OS function 'datetime.now' not implemented with standard execution
```

Today `datetime.now()` is only reachable via the external-call/resume path (see `crates/monty/test_cases/datetime__core.py` with `# call-external`). Reading the host clock should work in standard execution so common scripts run without requiring the host to plumb a suspend/resume loop.

nurikk:
## Summary
- Fixes pydantic/monty#330: `datetime.now()` previously raised `NotImplementedError: OS function 'datetime.now' not implemented with standard execution`.
- Standard (non-iterative) execution now satisfies `OsCall::DateTimeNow` and `OsCall::DateToday` by reading the host wall-clock directly, instead of falling through to the unimplemented path.
- Iterative `start`/`resume` callers are untouched — hosts can still inject a deterministic clock via `RunProgress::OsCall`.

> Draft: opening for upstream visibility. Note overlap in scope with #332, which takes a callback-based approach via `AbstractOS`. Happy to adapt or close in favor of #332.

## Approach
- New `host_date_today()` / `host_datetime_now(tz)` helpers in `crates/monty/src/os.rs` use `chrono` to produce `MontyObject::Date` / `MontyObject::DateTime`. Aware tz mirrors the existing test harness pattern (UTC shifted by fixed offset) and preserves the caller's `tz` so `datetime.tzinfo is tz` holds.
- `Executor::run_to_completion` in `crates/monty/src/run.rs` intercepts only the clock variants of `FrameExit::OsCall`; filesystem/env OS calls still surface as `NotImplementedError`, preserving the sandbox model.

## Test plan
- [x] New `crates/monty/test_cases/datetime__now_standard.py` covers naive, UTC, fixed-offset, `tz=` kwarg, `tz=None`, and `date.today()` — passes on Monty and CPython.
- [x] Pre-fix RED verified: same test failed with the original error.
- [x] Full datatest suite: 930 passed.
- [x] `cargo test -p monty --features ref-count-panic` passes (incl. iterative `os_tests.rs`).
- [x] `make pytest`: 891 passed.
- [x] `cargo clippy ... -D warnings`, `cargo fmt --check`, `make lint-py` all clean.

### DIFF
--- a/crates/monty/src/os.rs
+++ b/crates/monty/src/os.rs
@@ -10,7 +10,11 @@
 //! I/O, filesystem, or network operations. Instead, the host decides whether to
 //! permit and execute such operations.
 
-use crate::{ExcType, MontyException, MontyObject, intern::StaticStrings, types::str::StringRepr};
+use chrono::{Datelike, Local, Timelike, Utc};
+
+use crate::{
+    ExcType, MontyDate, MontyDateTime, MontyException, MontyObject, intern::StaticStrings, types::str::StringRepr,
+};
 
 /// OS operations that require host system access.
 ///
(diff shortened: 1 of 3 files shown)


samuelcolvin:
I would just fix this for the CLI as you originally suggested in #330.

samuelcolvin:
closing as no response.

Issue #335, opened 2026-04-15
Add `OsCall` Support for Symlink and Metadata Filesystem Operations


chaliy:
I run Monty against virtual file system and it actually working, I am looking for couple of optimizations, and would benefit from couple of new  OsCalls for:

- `Path.readlink`
- `Path.symlink_to`
- `Path.chmod`
- `Path.lstat` or `Path.stat(..., follow_symlinks=False)`

## Why These Would Help

`Path.readlink` - Monty can currently check whether a path is a symlink, but it cannot retrieve the symlink target.

This is a real limitation for embedded VFS implementations, where symlinks may be modeled explicitly and need to round-trip through Python APIs.

`Path.symlink_to` - Monty can observe symlinks but cannot create them.

Adding symlink creation would make the filesystem API more complete and better aligned with virtual filesystem backends that support symlink nodes.

`Path.chmod` - Monty already exposes mode bits through `stat`, but there is no way to modify them.

Even if some backends ignore or partially emulate permissions, it is useful to have a standard API surface so VFS implementations can decide how to support it.

`Path.lstat` or `follow_symlinks=False` - Once symlink support expands, non-following metadata becomes important too.

A separate `lstat`-style operation or a `follow_symlinks` flag on `stat` would let Monty distinguish between metadata for the link itself and metadata for its target.

## Suggested Implementation

Possible new `OsFunction` variants:

- `ReadLink`
- `Symlink`
- `Chmod`
- `Lstat` or a `Stat` request extended with `follow_symlinks`

Corresponding `pathlib.Path` methods:

- `Path.readlink()`
- `Path.symlink_to(target)`
- `Path.chmod(mode)`
- `Path.lstat()` or `Path.stat(follow_symlinks=False)`

## Non-goals

This issue is not proposing full file-object support such as:

- `open()`
- read/write streams
- append mode
- seek/tell
- context-managed file objects

Those are larger API areas and can be tackled separately.

## Open Questions

- Is `Path.lstat()` preferable to extending `Path.stat()` with `follow_symlinks`?
- Should `Path.symlink_to()` be implemented before `readlink()`, or vice versa?
- Should backends be allowed to raise `NotImplementedError` for `chmod` when permissions are not meaningful?

PR #336, opened 2026-04-15
Add pathlib symlink and metadata OS calls

chaliy:
## What changed

This adds the pathlib symlink and metadata filesystem operations requested in issue #335.

- add core OS-call support for `Path.readlink()`
- add core OS-call support for `Path.symlink_to(...)`
- add core OS-call support for `Path.chmod(...)`
- add core OS-call support for `Path.lstat()` and `Path.stat(follow_symlinks=False)`
- extend the Python `AbstractOS` / `OSAccess` surface accordingly
- add Rust and Python coverage for the new calls

## Why

Monty already exposed parts of pathlib filesystem access, but it could not round-trip symlink targets, create symlinks, change mode bits, or request non-following stat metadata. This closes that gap for the OS-call surface and the Python callback API.

## Notes

- direct/read-write mount handling supports the new operations
- `AbstractOS` remains backward-compatible via the `path_lstat()` fallback to `path_stat()`
- `OverlayMemory` still reports `Path.chmod()` and `Path.symlink_to()` as unsupported; that mode needs an explicit in-memory symlink model for full parity

## Validation

- `make format-rs`
- `make lint-rs`
- `make lint-py`
- `cargo test -p monty --test os_tests --features ref-count-panic`
- `cargo test -p monty --test fs --features ref-count-panic`
- `uv run pytest crates/monty-python/tests/test_os_calls.py crates/monty-python/tests/test_os_access.py crates/monty-python/tests/test_os_access_raw.py

### DIFF
--- a/crates/monty-datatest/src/main.rs
+++ b/crates/monty-datatest/src/main.rs
@@ -906,6 +906,9 @@ fn dispatch_os_call(
             // Virtual filesystem doesn't have symlinks
             MontyObject::Bool(false).into()
         }
+        OsFunction::Readlink => {
+            MontyException::new(ExcType::OSError, Some(format!("[Errno 22] Invalid argument: '{path}'"))).into()
+        }
         OsFunction::ReadText => {
             if let Some(file) = get_virtual_file(&path) {
                 match str::from_utf8(&file.content) {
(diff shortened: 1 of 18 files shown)


chaliy:
PR touches the monty-dataset, which triggers a considerable number of code paths that are not covered by unit tests. Should I still attempt to satisfy codecov?

samuelcolvin:
sorry this got forgotten, don't worry about codecov, but I don't think we should try to honor symlinks from the host fully, symlinks in `monty-fs` get very complicated. I would rather we returned something face to avoid errors than tried to accurately represent the real environment.

Closing this as the code has moved, so this will need to be restarted anyway.

PR #371, opened 2026-04-21
fixes GC interval ignored in ResourceTracker and Heap

sathish-t:
So a custom GC interval could be set in Python or JS but it was being ignored in favour of a default 100_000. I fixed this with a coding agent and approved its output. Basically, the `ResourceTracker` trait definition now gets a `fn gc_interval(&self) -> Option<usize>` and I implemented this for `PySignalTracker`,  `NoLimitTracker` and `LimitedTracker`. Then, the `Heap` 's `should_gc` now checks the tracker's `gc_interval()` and fires at a suitable interval. I also wrote some test infrastructure to check if the passed parameter is being respected.

 ## Tests passed

   - `make format-rs`
   - `make format-js`
   - `make lint-rs`
   - `make lint-js`
   - `make lint-py`
   - `cargo test -p monty --test resource_limits --features ref-count-return`
   - `make test-js`

### DIFF
--- a/crates/monty/src/heap.rs
+++ b/crates/monty/src/heap.rs
@@ -715,7 +715,7 @@ impl<'de, T: ResourceTracker + serde::Deserialize<'de>> serde::Deserialize<'de> 
 ///
 /// This is intentionally infrequent to minimize overhead while still
 /// eventually collecting reference cycles.
-const GC_INTERVAL: usize = 100_000;
+const DEFAULT_GC_INTERVAL: usize = 100_000;
 
 impl<T: ResourceTracker> Heap<T> {
     /// Creates a new heap with the given resource tracker.
@@ -1174,7 +1174,7 @@ impl<T: ResourceTracker> Heap<T> {
     /// and the number of allocations since the last GC exceeds the interval.
     #[inline]
     pub fn should_gc(&self) -> bool {
-        let interval = self.tracker.gc_interval().unwrap_or(GC_INTERVAL);
+        let interval = self.tracker.gc_interval().unwrap_or(DEFAULT_GC_INTERVAL);
         self.may_have_cycles.get() && (self.allocations_since_gc.get() as usize) >= interval
     }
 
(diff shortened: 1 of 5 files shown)


crates/monty/src/heap.rs
```diff
@@ -710,11 +710,12 @@ impl<'de, T: ResourceTracker + serde::Deserialize<'de>> serde::Deserialize<'de>
...
 }
 
-/// GC interval - run GC every 100,000 applicable allocations.
+/// Default GC interval - run GC every 100,000 applicable allocations unless
+/// the configured resource tracker overrides it.
 ///
 /// This is intentionally infrequent to minimize overhead while still
 /// eventually collecting reference cycles.
-const GC_INTERVAL: u32 = 100_000;
+const GC_INTERVAL: usize = 100_000;
```
davidhewitt:
```suggestion
const DEFAULT_GC_INTERVAL: usize = 100_000;
```

crates/monty/tests/resource_limits.rs
davidhewitt, line 1: There's another test in this file `gc_interval_triggers_collection` which looks totally broken, would be great to please fix this at the same time.

davidhewitt (commented):
Thanks, overall looks great, a couple of small suggestions

crates/monty/src/heap.rs
sathish-t: I've done this.

crates/monty/tests/resource_limits.rs
sathish-t: I've done it, I now do a test like the following, and check that allocations since last garbage collection is lower than the default 100,000, and the heap count is smaller than 20,000, and the success message is "done". Please let me know if this is a good test! I am a novice when it comes to these internal Python software concepts.

```python
result = 'done'
for i in range(210000):
    a = []
    a.append(a)
result
```

davidhewitt (approved):
Perfect, thanks very much!

PR #382, opened 2026-04-22
fix panic on duplicate parameter names

### LINKED ISSUE #377: Panic on function call when parameters have duplicate names
`pydantic-monty==0.0.16` parses `def f(x, x):` (and kw-only / mixed positional+kw-only variants) without error. CPython rejects the same source at compile time with `SyntaxError: duplicate argument 'x' in function definition`. The AST-level acceptance comes from Ruff's parser.

At call time the VM panics instead of raising a runtime error:

```
thread '<unnamed>' (NNNN) panicked at crates/monty/src/bytecode/vm/mod.rs:1758:32:
index out of bounds: the len is 1 but the index is 1
```

## Repro

```python
import pydantic_monty
pydantic_monty.Monty("def f(x, x):\n    return x\nprint(f(1, 2))").run()
```

Verified locally against `pydantic-monty==0.0.16`, Python 3.12.4. Also triggers on:

- `def f(*, x, x): return x` called as `f(x=1)`
- `def f(x, *, x=1): return x` called as `f(2)`

Both produce the same panic at `mod.rs:1758:32`.

## Expected

Either parse-time or call-time `SyntaxError` / `TypeError`, matching CPython. The panic path should not be reachable from user code.

## Likely cause

The function-signature binding appears to size the namespace slot array from unique parameter names. With a duplicate, the argument count exceeds the slot count and the index access at `mod.rs:1758` runs off the end.

Two possible fixes:

- Reject duplicate parameter names at Monty construction time, matching CPython's compile-time check.
- If Ruff's AST is treated as source of truth, add a VM-level check that raises `MontyTypingError` / `MontyRuntimeError` before the index access.

The parser-level acceptance is part of a known upstream umbrella in Ruff's tracker: [astral-sh/ruff#17412](https://github.com/astral-sh/ruff/issues/17412) covers duplicate parameter names and other semantic syntax errors Ruff still accepts. This issue is about the VM side - the index access at `mod.rs:1758` should not reach a panic regardless of what the parser admits.

smigolsmigol:
Fixes #377.

Ruff accepts `def f(x, x): return x` (and kw-only / mixed / varargs / kwargs / pos-only / lambda variants) while CPython rejects all of them at compile time. Monty's `Prepare::new_function` builds `name_map` as an `AHashMap`, so duplicates collapse to one entry; `namespace_size` is sourced from `name_map.len()` but each `NamespaceId` comes from the positional `enumerate()` index. With any duplicate name the resolved `NamespaceId` points past the allocated frame, and `load_local` at `vm/mod.rs` indexes past the end of the stack region and panics.

This adds `reject_duplicate_params` and calls it from `prepare_function_def` and `prepare_lambda`. The message matches CPython's `duplicate argument '{name}' in function definition` verbatim.

## Tests

- `parse_errors::duplicate_positional_parameter_returns_syntax_error`
- `parse_errors::duplicate_keyword_only_parameter_returns_syntax_error`
- `parse_errors::duplicate_mixed_positional_and_keyword_only_parameter_returns_syntax_error`
- `parse_errors::duplicate_lambda_parameter_returns_syntax_error`
- `crates/monty/test_cases/function__err_duplicate_param.py` (datatest harness, `# Raise=` format)

Also verified end-to-end against `monty-cli` on: positional dup, kw-only dup, mixed, lambda, `*args`/positional collision, `**kwargs`/positional collision, pos-only/positional collision, `*args`/`**kwargs` collision, triple dup, multi-collision. All raise the same CPython-matching `SyntaxError`.

## Caret position

CPython points at the duplicate parameter's position; the error points at the function name (`name.position` for `def`, the lambda `position` for `lambda`). Parameter-level position isn't threaded through `ParsedParam`. Sharper caret would need adding `position: CodeRange` there. Separate refactor.

## Not covered

`def f[T, T](...)` (duplicate type parameters) is a separate code path not addressed here. CPython raises `SyntaxError: duplicate type parameter 'T'`; Monty accepts it silently. Separate ticket.

### DIFF
--- a/crates/monty/src/prepare.rs
+++ b/crates/monty/src/prepare.rs
@@ -249,10 +249,7 @@ impl<'i> Prepare<'i> {
         let mut name_map = AHashMap::with_capacity(capacity);
         for (index, string_id) in params.iter().enumerate() {
             let name_str = interner.get_str(*string_id);
-            if name_map
-                .insert(name_str.to_string(), NamespaceId::new(index))
-                .is_some()
-            {
+            if name_map.insert(name_str.to_string(), NamespaceId::new(index)).is_some() {
                 return Err(ParseError::syntax(
                     format!("duplicate argument '{name_str}' in function definition"),
                     position,
--- a/crates/monty/src/prepare.rs
+++ b/crates/monty/src/prepare.rs
@@ -230,6 +230,7 @@ impl<'i> Prepare<'i> {
     fn new_function(
         capacity: usize,
         params: &[StringId],
+        position: CodeRange,
         assigned_names: AHashSet<String>,
         global_names: AHashSet<String>,
         nonlocal_names: AHashSet<String>,
@@ -238,10 +239,25 @@ impl<'i> Prepare<'i> {
         enclosing_locals: Option<AHashSet<String>>,
         cell_var_names: AHashSet<String>,
         interner: &'i InternerBuilder,
-    ) -> Self {
+    ) -> Result<Self, ParseError> {
+        // Reject duplicate parameter names while building the name_map.
+        // Ruff's parser accepts `def f(x, x)` that CPython rejects at compile
+        // time; without this check, `name_map` is deduplicated by HashMap
+        // semantics but each `NamespaceId` comes from the positional index,
+        // so the duplicate slot lands past the allocated stack region and
+        // panics `load_local` at runtime.
         let mut name_map = AHashMap::with_capacity(capacity);
         for (index, string_id) in params.iter().enumerate() {
-            name_map.insert(interner.get_str(*string_id).to_string(), NamespaceId::new(index));
+            let name_str = interner.get_str(*string_id);
+            if name_map
+                .insert(name_str.to_string(), NamespaceId::new(index))
+                .is_some()
+            {
+                return Err(ParseError::syntax(
+                    format!("duplicate argument '{name_str}' in function definition"),
+                    position,
+                ));
+            }
         }
         let namespace_size = name_map.len();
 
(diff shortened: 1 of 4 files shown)


davidhewitt (changes requested):
Thanks for the PR. Rather than have a `reject_duplicate_params` function we should make `Prepare::new_function` reject duplicate parameters when it's building the map.

crates/monty/src/prepare.rs
```diff
@@ -1267,7 +1280,11 @@ impl<'i> Prepare<'i> {
         // Register the function name in the current scope
         let (name, _) = self.get_id(name);
 
-        // Extract param names from the parsed signature for scope analysis
+        // Extract param names from the parsed signature for scope analysis.
+        // Duplicate names (`def f(x, x)`) are rejected inside `new_function`
+        // when it builds the namespace `name_map`, since duplicates would
+        // desynchronize `name_map.len()` from the positional `NamespaceId`
+        // and later panic `load_local` at runtime.
```
davidhewitt:
```suggestion
        // Extract param names from the parsed signature for scope analysis
```

crates/monty/src/prepare.rs
```diff
@@ -1483,7 +1501,8 @@ impl<'i> Prepare<'i> {
         let body_as_node: ParseNode = Node::Return(body.clone());
         let body_nodes = vec![body_as_node];
 
-        // Extract param names from the parsed signature for scope analysis
+        // Extract param names from the parsed signature for scope analysis.
+        // Duplicates are rejected inside `new_function` (see its doc).
```
davidhewitt:
```suggestion
        // Extract param names from the parsed signature for scope analysis
```

davidhewitt (approved):
Thanks - will just reset some LLM-style comment churn and merge.

PR #386, opened 2026-04-22
Fix async await exception propagation

### LINKED ISSUE #323: External async functions exceptions are not correctly caught inside the sandbox
Hi, this is a terrific package, thank you very much for your work.

This is a bug I have noticed with the latest version: pydantic-monty v0.0.11.

When an external function raises an exception, try/except inside the Monty sandbox does not behave correctly for async functions: the exception escapes entirely to the host.

The try/except is bypassed and a MontyRuntimeError is raised at the host level:

```python
async def async_fail():
    raise ValueError("async boom")

m = pydantic_monty.Monty("""
async def run():
    try:
        await async_fail()
        return 'no error'
    except Exception as e:
        return 'caught: ' + str(e)

await run()
""", inputs=[])

r = await m.run_async(external_functions={"async_fail": async_fail})
# Raises MontyRuntimeError instead of returning 'caught: async boom'  ❌
```

---
Expected behavior: 'caught: async boom'.

tusharhqq:
## Summary
- Fixes #323.
- Route failed awaited futures back through normal VM exception handling so `try`/`except` around direct `await` expressions can catch them.
- Preserve the original exception for `asyncio.gather(...)` failures instead of surfacing sibling task cancellation.
- Add regression coverage for direct future failure, direct gather failure, and gathered coroutine failure.

## Testing
- `PATH="$HOME/.cargo/bin:$PATH" make format-rs`
- `make dev-py`
- `uv run pytest crates/monty-python/tests/test_start.py::test_future_snapshot_resume_exception_caught_at_await crates/monty-python/tests/test_start.py::test_future_snapshot_resume_gather_exception_caught_at_await crates/monty-python/tests/test_start.py::test_future_snapshot_resume_gathered_coroutine_exception_caught_at_await -q`
- `uv run pytest crates/monty-python/tests/test_start.py -q`
- `make lint-rs`
- `make lint-py`
- `uv run pytest crates/monty-python/tests/test_async.py -q`

### DIFF
--- a/crates/monty-python/tests/test_start.py
+++ b/crates/monty-python/tests/test_start.py
@@ -1218,6 +1218,30 @@ def test_future_snapshot_preserved_on_invalid_results():
     assert final.output == snapshot(42)
 
 
+def test_future_snapshot_resume_exception_caught_at_await():
+    """Exceptions used to resume a future are raised at the `await` expression."""
+    code = """
+try:
+    await fetch()
+    result = 'not caught'
+except ValueError as exc:
+    result = 'caught: ' + str(exc)
+
+result
+"""
+    progress = Monty(code).start()
+    assert isinstance(progress, pydantic_monty.FunctionSnapshot)
+    assert progress.function_name == snapshot('fetch')
+    call_id = progress.call_id
+
+    progress = progress.resume({'future': ...})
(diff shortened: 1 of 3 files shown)


tusharhqq:
i will complete this pr by tomorrow, i am still working on it, i needed feedback from codex and devin, that's why i created the pr

PR #390, opened 2026-04-23
fix: parse errors must not leak ruff AST Debug formatting

smigolsmigol:
`parse_identifier` and `parse_unpack_target_impl` format the offending AST node through Rust `Debug` when rejecting a target. The resulting error message is the full Debug representation of the `ruff_python_ast` node - struct name, `node_index`, `range`, nested `value`, `ctx`, etc. Anything that surfaces `MontyException::message()` reads that text verbatim.

## What the error looks like today

```python
>>> pydantic_monty.Monty("*a = [1, 2]").run()
# MontySyntaxError: Expected name, got Starred(ExprStarred { node_index: NodeIndex(None),
#   range: 0..2, value: Name(ExprName { node_index: NodeIndex(None),
#   range: 1..2, id: Name("a"), ctx: Store }), ctx: Store })
```

Same shape for `*x.y = 1`, `*x[0] = 1`, `for x.y in [1]: pass`. Each dumps the full nested AST struct instead of a human description.

## Fix

Adds `describe_expr_kind(&AstExpr) -> &'static str`: a match over all 33 `AstExpr` variants returning a short human description (`"starred expression"`, `"attribute"`, `"subscript"`, `"list comprehension"`, and so on). The two call sites format this string instead of the Debug of the node.

```
before: "Expected name, got Starred(ExprStarred { node_index: ... })"
after:  "Expected name, got starred expression"

before: "invalid unpacking target: Attribute(ExprAttribute { ... })"
after:  "invalid unpacking target: attribute"
```

Exhaustive match, no catch-all: if ruff adds a new `Expr` variant, the compiler forces an update instead of silently falling back to Debug.

## Tests

New in `crates/monty/tests/parse_errors.rs`, one per trigger shape:

- `starred_name_target_has_clean_message` - `*a = [1, 2]`
- `starred_attribute_target_has_clean_message` - `*x.y = 1`
- `starred_subscript_target_has_clean_message` - `*x[0] = 1`
- `for_loop_attribute_target_has_clean_message` - `for x.y in [1]: pass`

Each snapshots the full error message via `insta::assert_snapshot!` so any regression that reintroduces Debug formatting of AST nodes (struct names, `node_index`, `range`, `ctx: Store`, etc.) fails the snapshot diff loudly.

## Not covered

`for x.y in [1]: pass` is valid Python. CPython 3.11 through 3.14 accept it. Monty rejects it at `parse_unpack_target_impl` because attribute / subscript targets are not implemented for for-loops. That rejection is a separate issue; this PR only cleans up the error message emitted along the existing path. Separate ticket.

CPython's own messages (`"starred assignment target must be in a list or tuple"`, etc.) are not adopted verbatim here to keep the diff minimal. Switching to CPython prose verbatim is a separate follow-up.

### DIFF
--- a/crates/monty/src/parse.rs
+++ b/crates/monty/src/parse.rs
@@ -1249,13 +1249,10 @@ impl<'a> Parser<'a> {
     fn parse_identifier(&mut self, ast: AstExpr) -> Result<Identifier, ParseError> {
         match ast {
             AstExpr::Name(ast::ExprName { id, range, .. }) => Ok(self.identifier(&id, range)),
-            other => {
-                let kind = describe_expr_kind(&other);
-                Err(ParseError::syntax(
-                    format!("Expected name, got {kind}"),
-                    self.convert_range(other.range()),
-                ))
-            }
+            other => Err(ParseError::syntax(
+                format!("Expected name, got {}", describe_expr_kind(&other)),
+                self.convert_range(other.range()),
+            )),
         }
     }
 
@@ -1353,13 +1350,10 @@ impl<'a> Parser<'a> {
                 }
                 Ok(UnpackTarget::Tuple { targets, position })
             }
-            other => {
-                let kind = describe_expr_kind(&other);
-                Err(ParseError::syntax(
-                    format!("invalid unpacking target: {kind}"),
-                    self.convert_range(other.range()),
-                ))
-            }
+            other => Err(ParseError::syntax(
+                format!("invalid unpacking target: {}", describe_expr_kind(&other)),
+                self.convert_range(other.range()),
+            )),
         }
     }
 
--- a/crates/monty/src/parse.rs
+++ b/crates/monty/src/parse.rs
@@ -171,48 +171,6 @@ pub struct Parser<'a> {
     depth_remaining: u16,
 }
 
-/// Short human-readable name for an `AstExpr` variant, for use in
-/// user-facing parse errors. Avoids the Rust `Debug` formatting of the
-/// node, which would leak internal field names, ranges, and struct
-/// layout of `ruff_python_ast` into the HTTP error body.
-fn describe_expr_kind(expr: &AstExpr) -> &'static str {
-    match expr {
-        AstExpr::Name(_) => "name",
-        AstExpr::Starred(_) => "starred expression",
-        AstExpr::Attribute(_) => "attribute",
-        AstExpr::Subscript(_) => "subscript",
-        AstExpr::Call(_) => "function call",
-        AstExpr::Tuple(_) => "tuple",
-        AstExpr::List(_) => "list",
-        AstExpr::Set(_) => "set",
-        AstExpr::Dict(_) => "dict",
-        AstExpr::NumberLiteral(_) => "number literal",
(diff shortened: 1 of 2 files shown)


crates/monty/src/parse.rs
```diff
@@ -1249,10 +1249,13 @@ impl<'a> Parser<'a> {
...
         match ast {
             AstExpr::Name(ast::ExprName { id, range, .. }) => Ok(self.identifier(&id, range)),
-            other => Err(ParseError::syntax(
-                format!("Expected name, got {other:?}"),
-                self.convert_range(other.range()),
-            )),
+            other => {
+                let kind = describe_expr_kind(&other);
+                Err(ParseError::syntax(
+                    format!("Expected name, got {kind}"),
```
samuelcolvin:
```suggestion
                Err(ParseError::syntax(
                    format!("Expected name, got {}", describe_expr_kind(&other);),
```

crates/monty/src/parse.rs
```diff
@@ -1350,10 +1353,13 @@ impl<'a> Parser<'a> {
...
                 Ok(UnpackTarget::Tuple { targets, position })
             }
-            other => Err(ParseError::syntax(
-                format!("invalid unpacking target: {other:?}"),
-                self.convert_range(other.range()),
-            )),
+            other => {
+                let kind = describe_expr_kind(&other);
+                Err(ParseError::syntax(
+                    format!("invalid unpacking target: {kind}"),
```
samuelcolvin:
```suggestion
                Err(ParseError::syntax(
                    format!("invalid unpacking target: {}", describe_expr_kind(&other)),
```

crates/monty/tests/parse_errors.rs
```diff
@@ -450,3 +450,82 @@ fn long_source_line_does_not_overflow_column() {
...
+//
+// These guard the previously observed info-disclosure where compile-time
+// parse errors returned the Rust `Debug` formatting of the offending AST
+// node verbatim in the HTTP response body (`ExprStarred { node_index:
+// NodeIndex(None), range: ..., ... }`). The messages should now be short
+// human prose that does not expose ruff's internal struct layout.
+
+/// Asserts an error message is present, contains the expected prose,
+/// and does not leak any telltale `ruff_python_ast` Debug tokens.
+fn assert_no_debug_leak(msg: Option<&str>, expected_substr: &str) {
```
samuelcolvin: I would remove this and instead use insta to check the full error text.

samuelcolvin (commented):
overall LGTM. a few minor points, more generally I would use insta for tests rather than a custom test function.

smigolsmigol:
Addressed all three points :

inlined the `describe_expr_kind` calls at both sites, switched the 4 error-message tests to `insta::assert_snapshot!` with a top-level `use insta::assert_snapshot;` import to match the convention from #375, and removed the custom helper. 
Also renamed the first test to `starred_name_target_*` for consistency with the sibling `starred_attribute_target_*` / `starred_subscript_target_*` names. 42/42 parse_errors tests pass locally, fmt + clippy clean. 
 
Thanks for the review.

davidhewitt (approved):
Thanks

PR #397, opened 2026-04-26
fix GC tracking for `datetime.tzinfo`

### LINKED ISSUE #396: Panic in `Heap::inc_ref` (use-after-free) when GC sweeps `datetime.tzinfo` while the datetime still references it
## Summary

Aware `datetime` objects retain a private heap reference to their `tzinfo`
(`DateTime.tzinfo_ref`) so that `dt.tzinfo is tz` identity holds across
attribute lookups. However, the mark-and-sweep garbage collector does not
follow that reference, so a cycle-triggered collection can sweep the
`TimeZone` while the `DateTime` still points at the freed slot. The next
`dt.tzinfo` access panics in `Heap::inc_ref` with
`HeapEntries::get - data already freed`.

`pydantic_monty` version: latest `main` (`33e47c6`)

## Reproducer

```python
from datetime import datetime, timezone, timedelta

tz = timezone(timedelta(hours=5))
dt = datetime(2024, 1, 1, tzinfo=tz)
tz = None  # only `dt` keeps the timezone alive now

# Trigger cycle-detecting GC: a self-referential list flips
# `may_have_cycles=true`, and crossing GC_INTERVAL (100_000 GC-tracked
# allocations) runs a mark-and-sweep pass.
seed = []
seed.append(seed)
for i in range(100_100):
    x = [i]

# After GC the tzinfo's heap slot has been freed, but `dt` still points at it.
str(dt.tzinfo)
```

### Expected

`UTC+05:00` (the same as CPython prints).

### Actual

```
thread 'main' panicked at crates/monty/src/heap.rs:938:34:
HeapEntries::get - data already freed
```

Full backtrace (relevant frames):

```
5: monty::heap::heap_entries::HeapEntries::get
     at ./crates/monty/src/heap/heap_entries.rs:110:42
6: monty::heap::Heap<T>::inc_ref
     at ./crates/monty/src/heap.rs:927:34
7: monty::types::datetime::<impl PyTrait for HeapRead<DateTime>>::py_getattr
     at ./crates/monty/src/types/datetime.rs:1149:29
8: monty::heap_data::<impl PyTrait for HeapReadOutput>::py_getattr
     at ./crates/monty/src/heap_data.rs:914:38
9: monty::value::Value::py_getattr
     at ./crates/monty/src/value.rs:1692:67
```

## Root cause

Both `collect_child_ids` (GC mark traversal) and `py_dec_ref_ids_for_data`
(drop cascade) in `crates/monty/src/heap.rs` end with a `_ => {}` arm that
silently swallows `HeapData::DateTime`. So:

1. **GC mark phase** never visits the `tzinfo` via the `DateTime`. If no
   other GC root holds the timezone, it's swept while the datetime still
   references it.
2. **Drop cascade** never decrements the tzinfo's refcount when a
   datetime is freed (mirror bug, opposite direction, causing a refcount
   leak).

The defensive panic in `HeapEntries::get` catches the unsafe read so this
is a DoS bug rather than memory disclosure, but the underlying refcount
imbalance is real and reachable from user code.

## Fix

Add `HeapData::DateTime(dt)` arms to both functions that visit
`dt.tzinfo_ref`. Plus an accessor on `DateTime` and a regression test
under `datetime__core.py`. Will follow up with a PR.

rsr5:
Closes #396

## Summary

Aware `datetime` objects retain a heap reference to their `tzinfo`
(`DateTime.tzinfo_ref`) so that `dt.tzinfo is tz` identity holds across
attribute lookups. The mark-and-sweep GC was not following that reference,
so a cycle-triggered collection could sweep the timezone while the
datetime still pointed at the freed slot. The next `dt.tzinfo` access
panicked in `Heap::inc_ref` with `HeapEntries::get - data already freed`.

## Fix

- Add `HeapData::DateTime` arms to both `collect_child_ids` (GC mark) and
  `py_dec_ref_ids_for_data` (drop cascade) in `heap.rs`, each visiting
  `tzinfo_ref`.
- Expose `DateTime::tzinfo_ref()` as a `pub(crate)` accessor so `heap.rs`
  doesn't need to reach into the private field directly.
- Regression test added to `datetime__core.py`.

## Before

```
$ cargo run -p monty-cli -- repro.py
thread 'main' panicked at crates/monty/src/heap.rs:938:34:
HeapEntries::get - data already freed
```

## After

```
$ cargo run -p monty-cli -- repro.py
UTC+05:00
```

## Test plan

- [x] Original reproducer runs to completion and prints `UTC+05:00`
- [x] New regression test in `datetime__core.py` (GC-must-follow-tzinfo section) passes under `--features memory-model-checks`
- [x] All 50 `refcount__*` tests pass
- [x] All 8 `pyobject__cycle_*` tests pass
- [x] `cargo fmt` and `cargo clippy` clean

### DIFF
--- a/crates/monty/test_cases/datetime__core.py
+++ b/crates/monty/test_cases/datetime__core.py
@@ -1183,25 +1183,20 @@ except TypeError as e:
 
 
 # === GC must follow datetime.tzinfo_ref ===
-# Regression: an aware datetime retains the tzinfo as a private heap reference
-# (so `dt.tzinfo is tz` identity holds across attribute lookups). Without GC
-# tracking that reference, a cycle-triggered collection could sweep the
-# tzinfo while the datetime still pointed at the freed slot, panicking on
-# the next `dt.tzinfo` access with `HeapEntries::get - data already freed`.
+# Regression: aware datetimes retain the tzinfo as a private heap reference,
+# so the GC mark phase must follow it. Without that, a cycle-triggered
+# collection sweeps the tzinfo while the datetime still points at the freed
+# slot, causing the next `dt.tzinfo` access to either panic with
+# `HeapEntries::get - data already freed` or read whatever was reallocated
+# into the slot.
 tz_keepalive = datetime.timezone(datetime.timedelta(hours=5))
 dt_keepalive = datetime.datetime(2024, 1, 1, tzinfo=tz_keepalive)
 tz_keepalive = None  # only `dt_keepalive` keeps the timezone alive now
 
-# Force the cycle-detecting GC to run: a self-referential list flips
-# `may_have_cycles`, and crossing the GC_INTERVAL allocation threshold
-# (100_000 GC-tracked allocations) triggers a mark-and-sweep pass.
+# Flip `may_have_cycles=true` and trigger one allocation; under
+# `--features memory-model-checks` (CI default) GC fires on every alloc.
 gc_seed = []
 gc_seed.append(gc_seed)
-for _i in range(100_100):
-    _x = [_i]
+_ = []  # triggers GC
 
-# After GC, the tzinfo must still be reachable through the datetime.
 assert str(dt_keepalive.tzinfo) == 'UTC+05:00', 'datetime tzinfo must survive GC'
-assert dt_keepalive.tzinfo == datetime.timezone(datetime.timedelta(hours=5)), (
-    'datetime tzinfo must round-trip equal after GC'
-)
--- a/crates/monty/test_cases/datetime__core.py
+++ b/crates/monty/test_cases/datetime__core.py
@@ -1180,3 +1180,28 @@ try:
     assert False, 'datetime with duplicate month should raise TypeError'
 except TypeError as e:
     assert str(e) == "argument for function given by name ('month') and position (2)", f'dt dup month: {e}'
+
+
+# === GC must follow datetime.tzinfo_ref ===
+# Regression: an aware datetime retains the tzinfo as a private heap reference
+# (so `dt.tzinfo is tz` identity holds across attribute lookups). Without GC
+# tracking that reference, a cycle-triggered collection could sweep the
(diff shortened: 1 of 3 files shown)


crates/monty/test_cases/datetime__core.py
```diff
@@ -1180,3 +1180,28 @@
...
+dt_keepalive = datetime.datetime(2024, 1, 1, tzinfo=tz_keepalive)
+tz_keepalive = None  # only `dt_keepalive` keeps the timezone alive now
+
+# Force the cycle-detecting GC to run: a self-referential list flips
+# `may_have_cycles`, and crossing the GC_INTERVAL allocation threshold
+# (100_000 GC-tracked allocations) triggers a mark-and-sweep pass.
+gc_seed = []
+gc_seed.append(gc_seed)
+for _i in range(100_100):
+    _x = [_i]
```
samuelcolvin: @davidhewitt do we need this? I thought we set the GC limit to way lower for tests.

samuelcolvin (commented):
generally LGTM, but I'll let @davidhewitt review as he understands  the GC better than me.

crates/monty/test_cases/datetime__core.py
davidhewitt: With the `memory-model-checks` feature enabled the GC should be queued to run after every allocation.

davidhewitt (commented):
Thanks, this is a correct fix to me (maybe the test can be simplified).

Overall the need to have this kind of code all over the place to trace data through the GC seems like a real problem, I'm actively researching solutions.

rsr5:
Thanks @davidhewitt! Pushed a simpler regression test in e96e3c6. Under `--features memory-model-checks` the GC interval is 1, so a self-referential list plus one extra allocation is enough to trigger a collection. No need for the 100k-iteration loop. Verified it still fails without the fix.

davidhewitt (approved):
Thanks

PR #410, opened 2026-05-03
fix `no active frame` panic on duplicate coroutine in `gather`

### LINKED ISSUE #409: `no active frame` panic when `asyncio.gather` is given the same coroutine twice
## Repro

```python
import asyncio

async def f():
    return 1

c = f()
asyncio.run(asyncio.gather(c, c))
```

```
thread 'main' panicked at crates/monty/src/bytecode/vm/mod.rs:1628:28:
no active frame
```

Verified against `main` at `49faa4c`.

## Root cause

When `asyncio.gather` is given two references to the same coroutine, the scheduler spawns one task per argument. Each task's startup goes through `load_or_init_task` → `init_task_from_coroutine` (`crates/monty/src/bytecode/vm/async_exec.rs:506`, `:569`).

Task 1 starts, runs `f()` to completion, and pops its frame. The scheduler then loads task 2 with empty frames. `init_task_from_coroutine` (`async_exec.rs:574-579`) sees the coroutine is no longer in state `New` and returns:

```rust
return Err(
    SimpleException::new_msg(ExcType::RuntimeError, "cannot reuse already awaited coroutine").into(),
);
```

That `Err` propagates into `handle_exception`, which calls `self.current_frame()`. Task 2 has no frames at this point (task 1's were popped, task 2's were never pushed), so `current_frame()` panics on the empty stack.

## CPython behaviour

CPython doesn't raise — `asyncio.gather` treats both arguments as the same task and returns its result twice:

```python
>>> asyncio.run(asyncio.gather(c, c))
[1, 1]
```

### LINKED ISSUE #412: Panic in `asyncio.gather(...)` when the same external future is passed twice
## Repro

Guest code:

```python
import asyncio

f = source()
vals = await asyncio.gather(f, f)
vals
```

Host-side callback used for the external function:

```python
async def source():
    return 1
```

The public resume path panics with:

```text
pyo3_runtime.PanicException: all results should be filled when gather is complete
```

## Likely cause

Passing the same external future twice appears to alias one pending future into two conceptual gather result slots. When that future resolves, gather completion is reached with only one slot filled, which trips the invariant in the result collection path.

Relevant source:

- `crates/monty/src/bytecode/vm/async_exec.rs:705-707`

## Related issues

- Distinct from [#409](https://github.com/pydantic/monty/issues/409), which covers passing the same coroutine twice and panicking with `no active frame`.
- Related to [#240](https://github.com/pydantic/monty/issues/240) / [#251](https://github.com/pydantic/monty/pull/251), but this trigger is duplicate aliasing of one external future rather than partial resolution of a mixed gather.

rsr5:
Fixes #409
Fixes #412

## What this changes

`handle_exception` and `current_frame_name` no longer assume `self.frames` is non-empty. When an exception is raised in the no-frames window (currently only reachable via `init_task_from_coroutine` for an already-awaited coroutine), `current_frame_name` falls back to `<module>` and `handle_exception` either routes spawned-task failures through `handle_task_failure` or propagates up.

The repro from #409 now raises `RuntimeError: cannot reuse already awaited coroutine` instead of panicking.

Regression test added to `crates/monty/test_cases/async__gather_all.py`.

## A note on scope

This fixes the panic but introduces a CPython divergence: CPython's `asyncio.gather` dedupes coroutines and returns `[1, 1]` for the repro, whereas Monty now cleanly raises `RuntimeError`. The regression test accepts either behaviour so it does not lock that in.

Happy to look into a wider fix that aligns with CPython's dedup semantics if that is what you want, otherwise let me know what shape you would prefer.

## Test plan

- [x] `make test-memory-model-checks`
- [x] Repro from #409 no longer panics
- [x] `async__gather_all.py` passes under both Monty and CPython runners

### DIFF
--- a/crates/monty/src/bytecode/vm/async_exec.rs
+++ b/crates/monty/src/bytecode/vm/async_exec.rs
@@ -8,6 +8,8 @@
 
 use std::mem;
 
+use ahash::AHashMap;
+
 use super::{AwaitResult, CallFrame, FrameExit, VM};
 use crate::{
     MontyException,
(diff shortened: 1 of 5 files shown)


davidhewitt:
I think consistency with CPython semantics would be much better. Do you have a sense from working on this how hard that'd be?

rsr5:
I had some time to look at it and thought it would be easier to just fix it. Force-pushed; the original defensive guard is gone, replaced by two parts.

## CPython-compatible deduplication

`asyncio.gather` in CPython dedupes by object identity (`arg_to_fut` in `asyncio.tasks.gather`): the same coroutine passed multiple times runs once and its result is reused at every duplicate position. Monty now matches:

```python
async def f():
    return 1

c = f()
await asyncio.gather(c, c)  # [1, 1]; f() runs once
```

`Task.gather_result_idx` becomes a `Vec<usize>`. `await_gather_future` builds a per-unique spawn plan that preserves first-occurrence order; `handle_task_completion` fans the result across all the task's slots (clone for each but the last, move into the last). Same dedup applies to repeated external futures via the gather-waiter map, which incidentally fixes a pre-existing bug where `gather(f, f)` with an already-resolved external future would leave the second slot unfilled forever.

## Fix-at-source for the empty-frames panic

The dedup covers `gather(c, c)`, but the same panic was reachable via `await c; gather(c)` (no duplicates, just an already-awaited coroutine). `load_or_init_task` now pre-checks the coroutine state and routes "already awaited" failures through `handle_task_failure`, which restores the waiter's frames before the error propagates. The exception machinery never sees an empty frame stack, so the defensive guard from the previous revision is removed.

`await c; gather(c)` now raises `RuntimeError: cannot reuse already awaited coroutine` cleanly with full traceback, matching CPython.

## Tests

`async__gather_all.py` adds: `gather(c, c)` returns `[1, 1]` with body running exactly once (counter); 3-way and mixed `gather(a, b, a)` cases; `await c; gather(c)` raises the exact CPython error message. All assertions are exact and pass under both runners.

Tested clean under `memory-model-checks`, `ref-count-return`, no-features, and clippy.

davidhewitt (approved):
Thanks!

PR #413, opened 2026-05-05
fix: avoid panic when list.sort() key callback mutates the list

### LINKED ISSUE #411: Panic in `list.sort(key=...)` when a key callback mutates the list reentrantly
## Repro

```python
xs = [3, 2, 1]


def key(value):
    xs.clear()
    return value


xs.sort(key=key)
```

A shrink-by-one variant also reproduces:

```python
xs = [3, 2, 1]


def key(value):
    if len(xs) == 3:
        xs.pop()
    return value


xs.sort(key=key)
```

The first repro panics with:

```text
thread '<unnamed>' panicked at crates/monty/src/types/list.rs:736:47:
index out of bounds: the len is 0 but the index is 1
```

The second repro panics with:

```text
thread '<unnamed>' panicked at crates/monty/src/types/list.rs:736:47:
index out of bounds: the len is 2 but the index is 2
```

## Root cause

`do_list_sort` snapshots the original list length, then iterates `for i in 0..len` while user-controlled `key` callbacks are still allowed to mutate the same list. After a callback shrinks the list, the later `items[i]` access uses the stale loop bound and indexes past the new end.

Relevant source:

- `crates/monty/src/types/list.rs:729`
- `crates/monty/src/types/list.rs:735-737`

SAY-5:
Closes #411.

`do_list_sort` snapshotted the list's length and then iterated `for i in 0..len` while user-controlled `key` callbacks were still allowed to mutate the same list. Once a callback shrank the list, `items[i]` indexed past the new end and panicked.

This change follows CPython's strategy: detach the list's `items` vector for the duration of the sort so reentrant access via the list's heap id sees an empty list, run all key extraction / comparison / permutation work on the detached buffer, and afterwards either restore the sorted buffer (success) or - if the user repopulated the empty list during sort - drop their interleaved mutations, restore the (sorted-or-partially-permuted) buffer to keep refcount invariants intact, and raise `ValueError: list modified during sort` matching CPython exactly.

Regression tests added to `crates/monty/test_cases/list__ops.py` for the two repros from the issue (`clear()` and `pop()` in the key callback) plus the repopulation case that should raise `ValueError`.

### DIFF
--- a/crates/monty/src/exception_private.rs
+++ b/crates/monty/src/exception_private.rs
@@ -1230,6 +1230,18 @@ impl ExcType {
         SimpleException::new_msg(Self::ValueError, "list.remove(x): x not in list").into()
     }
 
+    /// Creates a ValueError for `list.sort()` when the list was reentrantly
+    /// mutated by a user-supplied `key` callback (or comparison) during the sort.
+    ///
+    /// CPython detaches the list's storage for the duration of the sort so the
+    /// list looks empty to reentrant code, then raises this error if the user
+    /// re-populated the list while the sort was in progress. Matches CPython's
+    /// exact message: `ValueError: list modified during sort`.
+    #[must_use]
+    pub(crate) fn value_error_list_modified_during_sort() -> RunError {
+        SimpleException::new_msg(Self::ValueError, "list modified during sort").into()
+    }
+
     /// Creates an IndexError for popping from an empty list.
     ///
     /// Matches CPython's format: `IndexError: pop from empty list`
--- a/crates/monty/src/exception_private.rs
+++ b/crates/monty/src/exception_private.rs
@@ -1230,18 +1230,6 @@ impl ExcType {
         SimpleException::new_msg(Self::ValueError, "list.remove(x): x not in list").into()
     }
 
-    /// Creates a ValueError for `list.sort()` when the list was reentrantly
-    /// mutated by a user-supplied `key` callback (or comparison) during the sort.
-    ///
-    /// CPython detaches the list's storage for the duration of the sort so the
-    /// list looks empty to reentrant code, then raises this error if the user
-    /// re-populated the list while the sort was in progress. Matches CPython's
-    /// exact message: `ValueError: list modified during sort`.
-    #[must_use]
-    pub(crate) fn value_error_list_modified_during_sort() -> RunError {
-        SimpleException::new_msg(Self::ValueError, "list modified during sort").into()
-    }
-
     /// Creates an IndexError for popping from an empty list.
     ///
     /// Matches CPython's format: `IndexError: pop from empty list`
--- a/crates/monty/src/types/list.rs
+++ b/crates/monty/src/types/list.rs
@@ -8,7 +8,7 @@ use crate::{
     args::ArgValues,
     bytecode::{CallResult, VM},
     defer_drop, defer_drop_mut,
-    exception_private::{ExcType, RunError, RunResult},
+    exception_private::{ExcType, RunError, RunResult, SimpleException},
     heap::{DropWithHeap, Heap, HeapData, HeapGuard, HeapId, HeapItem, HeapRead, HeapReadOutput, HeapReader},
     intern::StaticStrings,
     resource::{ResourceError, ResourceTracker},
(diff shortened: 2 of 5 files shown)


crates/monty/test_cases/list__ops.py
```diff
@@ -325,6 +325,56 @@ def last_char(s):
...
+
+
+def pop_key(value):
+    if xs2:
+        xs2.pop()
+    return value
+
+
+xs2.sort(key=pop_key)
+assert xs2 == [1, 2, 3], 'sort with key that pops from the list still produces sorted output'
```
davidhewitt: This test seems pointless to me, the answer is that the list appears empty during the sort so the pop never runs.

Perhaps a more interesting variant is a key which has `assert len(xs2) == 0`?

crates/monty/test_cases/list__ops.py
```diff
@@ -325,6 +325,56 @@ def last_char(s):
...
+# list, then raises ValueError if the user repopulated it. After a successful
+# sort the list is restored to the sorted order.
+
+# clear() during key extraction - list looks empty to the callback
+xs1 = [3, 2, 1]
+
+
+def clear_key(value):
+    xs1.clear()
+    return value
```
davidhewitt, line 339: CPython has a variant, if the key is

```python
def key(arg):
    a.append(1)
    a.clear()
    return arg
```

Then this will still lead to the `ValueError: list modified during sort`.

crates/monty/src/exception_private.rs
```diff
@@ -1230,6 +1230,18 @@ impl ExcType {
...
+    /// mutated by a user-supplied `key` callback (or comparison) during the sort.
+    ///
+    /// CPython detaches the list's storage for the duration of the sort so the
+    /// list looks empty to reentrant code, then raises this error if the user
+    /// re-populated the list while the sort was in progress. Matches CPython's
+    /// exact message: `ValueError: list modified during sort`.
+    #[must_use]
+    pub(crate) fn value_error_list_modified_during_sort() -> RunError {
+        SimpleException::new_msg(Self::ValueError, "list modified during sort").into()
+    }
```
davidhewitt: I see no point in having a free function for this, it has one caller, just inline it in `list.rs`.

crates/monty/src/types/list.rs
```diff
@@ -725,25 +737,73 @@ fn do_list_sort<'h>(
...
+        let keys: Vec<Value> = Vec::with_capacity(items.len());
         // Use a HeapGuard to ensure that if key function evaluation fails partway through,
         // we clean up any keys that were successfully computed
         let mut keys_guard = HeapGuard::new(keys, vm);
         let (keys, vm) = keys_guard.as_parts_mut();
-        for i in 0..len {
-            let item = list.get(vm.heap).items[i].clone_with_heap(vm);
+        // Iterate the detached buffer directly. Reentrant mutation via the
+        // list's heap id targets the empty live list, not this buffer, so we
+        // never observe a shifting length here.
```
davidhewitt: I don't think comment on "shifting length" is at all relevant here now that this operates on detached items.

```suggestion
```

crates/monty/src/types/list.rs
```diff
@@ -698,6 +698,18 @@ fn list_count<'h>(
...
 
 /// Performs an in-place sort on a list with optional key function and reverse flag.
+///
+/// To safely support user-supplied `key` callbacks (and rich-comparison `__lt__`
+/// methods) that may reentrantly mutate the same list, we follow CPython's
+/// strategy: the list's `items` vector is **detached** for the duration of the
+/// sort so the list looks empty to any reentrant code. All sort work is then
+/// performed on the detached buffer. After sorting, if the list is still empty
+/// we restore the (now sorted) buffer; if the user repopulated the list during
+/// the sort we discard our buffer and raise
```
davidhewitt: This seems wrong, we don't discard our buffer, we always restore it.

crates/monty/src/types/list.rs
```diff
@@ -698,6 +698,18 @@ fn list_count<'h>(
...
+/// `ValueError: list modified during sort`, matching CPython exactly.
+///
+/// This avoids the previous panic where a stale loop-bound captured before
+/// iteration could index past the live list's length after a callback shrank it.
```
davidhewitt: Comment describing previous state of the code is not useful.


```suggestion
```

crates/monty/src/types/list.rs
```diff
@@ -725,25 +737,73 @@ fn do_list_sort<'h>(
     };
     defer_drop!(key_fn, vm);
 
-    // 1. Compute key values if a key function was provided, otherwise we'll sort by the items themselves
+    // Detach the list's items so reentrant access via the list's heap id sees
+    // an empty list. The detached items still carry their existing refcounts;
+    // we'll either move them back into the list (success) or drop them
+    // (mutation detected / error during sort).
```
davidhewitt: We always restore the detached buffer.

crates/monty/src/types/list.rs
```diff
@@ -725,25 +737,73 @@ fn do_list_sort<'h>(
...
+    let was_modified = !list.get(vm.heap).items.is_empty();
+    if was_modified {
+        // Discard whatever the user appended/inserted on the empty live list.
+        mem::take(&mut list.get_mut(vm.heap).items).drop_with_heap(vm);
+    }
+    // Restore the detached buffer (sorted on success, partially-permuted on
+    // sort error). The `contains_refs` flag was not cleared by `mem::take`
+    // (only the Vec was taken), so it correctly still describes the buffer
+    // being put back.
+    *list.get_mut(vm.heap).as_vec_mut() = items;
```
davidhewitt: I think this can be simplified with a `std::mem::swap` to put the items back in, and then if the temp items are not empty we raise.

davidhewitt (commented):
Thanks for the PR, please see comments.

davidhewitt:
Sorry about that, `main` should be working again now. Please action the remaining comments from my review above.

davidhewitt (approved):
Thanks very much! I pushed a commit which does some deduplication now possible, and will merge.

PR #420, opened 2026-05-07
reject reusing an already-awaited external future via `gather`

rsr5:
While working on #410 I noticed a related case at the cross-gather boundary that #410's intra-gather dedup didn't cover.

Repro:

```py
async def helper(future):
    return await asyncio.gather(future)

shared = async_call(99)
await asyncio.gather(helper(shared), helper(shared))
```

The first `helper`'s gather registers as a waiter on `shared`; the second silently overwrites that registration in the scheduler's `gather_waiters` map. When `shared` resolves, only the second helper sees it and the first hangs forever.

This PR rejects the second use up front with the existing `RuntimeError: cannot reuse already awaited future`, matching how direct `await f; await f` already behaves. CPython raises the analogous `cannot reuse already awaited coroutine` at the same point.


davidhewitt (approved):
Thanks!

PR #421, opened 2026-05-07
implement exception `__context__`

davidhewitt:
This PR implements exception `__context__` attribute plus fixes various cases with bare `raise` that I (with Claude's help) found while doing so (see new test cases).


davidhewitt:
This is problematic because `__context__` introduces heap references into `RunError`. Need a strategy to get that right before we proceed here.

PR #424, opened 2026-05-09
Fix async external function exception propagation

### LINKED ISSUE #323: External async functions exceptions are not correctly caught inside the sandbox

vivekraghuram:
Resolves https://github.com/pydantic/monty/issues/323

Based somewhat on https://github.com/pydantic/monty/pull/386

Added tests to mirror the sync external function exception handling, plus a few async-specific tests for asyncio.gather, asyncio.run, and user-defined async wrappers.

The fix routes failed awaited futures through `resume_with_exception` instead of returning the error directly. That allows `try`/`except` around an `await` to catch the exception. Uncaught exceptions still propagate out to the host unchanged.

---
## Summary by cubic
Fixes exception propagation for awaited async external callbacks so Monty try/except can catch them; uncaught exceptions (including in `asyncio.gather` and nested coroutines) still bubble to the host unchanged. Fixes pydantic/monty#323.

- **Bug Fixes**
  - VM: route failures from awaited externals through `resume_with_exception(err)` instead of returning `Err(err)` so surrounding `try/except` can catch them; uncaught behavior is unchanged.
  - Datatest harness: add `DispatchResult::AsyncFail(MontyException)` and store pending futures as pre-built `ExtFunctionResult` so `ResolveFutures` handles both `Return` and `Error`; add and wire `async_fail` into name lookup.
  - Tests: add `crates/monty/test_cases/async__ext_exc.py` mirroring sync cases plus `asyncio.gather` and wrapper shapes; add `crates/monty-python/tests/test_external_async.py` for host `MontyRuntimeError` and return-conversion errors; expose `async_fail` in `scripts/iter_test_methods.py`.

<sup>Written for commit 848cc4f496afa4462469fa5f2e939317e84588e3. Summary will update on new commits.</sup>

### DIFF
--- a/crates/monty-python/tests/test_external_async.py
+++ b/crates/monty-python/tests/test_external_async.py
@@ -1,0 +1,60 @@
+"""Tests for the async external-function surface of the Python bindings."""
+
+import pytest
+from inline_snapshot import snapshot
+
+import pydantic_monty
+
+
+async def test_async_external_function_raises_surfaces_as_monty_runtime_error():
+    """An uncaught exception from an awaited async callback surfaces as
+    ``MontyRuntimeError`` with the original exception preserved in
+    ``exc.exception()``."""
+    m = pydantic_monty.Monty('await fail()')
+
+    async def fail():
+        raise ValueError('intentional error')
+
+    with pytest.raises(pydantic_monty.MontyRuntimeError) as exc_info:
+        await m.run_async(external_functions={'fail': fail})
+    inner = exc_info.value.exception()
+    assert isinstance(inner, ValueError)
+    assert inner.args[0] == snapshot('intentional error')
+
+
+async def test_async_external_function_return_lone_surrogate_catchable_inside_monty():
+    """An async callback returning a string with a lone surrogate surfaces inside Monty
+    as a ``ValueError`` that can be caught, not as a raw ``PyErr`` escaping to the caller."""
+    code = """
+try:
+    await get_str()
+    result = 'no error'
+except ValueError:
+    result = 'caught'
+result
+"""
+    m = pydantic_monty.Monty(code)
+
+    async def get_str():
+        return '\ud83d'
+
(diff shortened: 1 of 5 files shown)


crates/monty-python/tests/test_external_async.py
davidhewitt, line 1: We should move these tests into `crates/monty/test_cases`. You'll probably need to modify `crates/monty-datatest/src/main.rs` to have an `async_fail` external future (currently only has `async_call` which unconditionally succeeds). Or change `async_call` to have a second optional `raise=True` flag. Not sure what is easier.

davidhewitt (commented):
Thanks for the PR! I would like some changes for the tests, otherwise fix seems correct.

crates/monty-python/tests/test_external_async.py
vivekraghuram: Done! I added an async_fail case and moved most of the tests to `async__ext_exc.py`. I did leave a few in test_external_async.py cause I think they are useful and don't belong there but happy to drop these tests entirely if you think they're redundant.

davidhewitt (approved):
Thanks!

PR #434, opened 2026-05-14
Bound container pre-allocation driven by iterator size hints

dsp-ant:
Constructing a set, frozenset, or map result eagerly pre-allocates a container with capacity drawn from the iterator's size hint. The hint for an iterable like a large range is its full remaining length, so the pre-allocation can reach the global allocator with a request far beyond the configured memory budget. An allocation failure aborts the host rather than raising a catchable exception.

Add a preallocation helper on the iterator that validates the requested capacity against the resource tracker (raising MemoryError when it would exceed the budget) and clamps the result to a small bound so the pre-allocation is defensively safe even when no limits are configured.

---
## Summary by cubic
Prevents unbounded allocations from large iterables and dynamic f-string widths. All pre-allocation and collection are checked against the resource tracker; over-budget raises MemoryError instead of crashing.

- **Bug Fixes**
  - Added `MontyIter::preallocation_hint` and used it in set/frozenset/dict-view and `map()`; validates via `check_estimated_size` and clamps to 65,536.
  - Bounded `MontyIter::collect` incrementally so `list()`, `tuple()`, `sorted()`, `reversed()`, and generator builds stop near the limit.
  - Guarded f-string dynamic field width with `check_repeat_size` before padding; expanded resource-limit tests for these paths; small cases still succeed.

<sup>Written for commit 42c446950cf83a4847efd538849d413b87400e16. Summary will update on new commits. <a href="https://cubic.dev/pr/pydantic/monty/pull/434?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/crates/monty/src/builtins/map.rs
+++ b/crates/monty/src/builtins/map.rs
@@ -1,6 +1,6 @@
 //! Implementation of the map() builtin function.
 
-use std::iter;
+use std::{iter, mem};
 
 use crate::{
     args::{ArgValues, KwargsValues},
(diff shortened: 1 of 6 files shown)


dsp-ant-monty-findings-2026-05-08.md
davidhewitt, line 1: Possibly not intended to commit this?

davidhewitt (commented):
Thanks, I suspect there are probably a similar class off issues caused by calls to `.collect()` that don't first test size hints (or should check length during collection).

dsp-ant:
> Thanks, I suspect there are probably a similar class off issues caused by calls to `.collect()` that don't first test size hints (or should check length during collection).

Ah right. I had the Claude implement it inside MontyIter::collect now. What do you think? Happy to change it. I am not super familiar with the codebase and obv using claude to make it work.

davidhewitt (approved):
Let's do this, thanks 👍

PR #435, opened 2026-05-14
Pre-check requested size in bytes(int) constructor against resource l…

dsp-ant:
The integer-argument form of bytes() allocated a zero-filled buffer proportional to the supplied count without consulting the resource tracker first. Add the same check_repeat_size guard used by the sibling padding/repeat operations so a configured memory limit raises MemoryError before the native allocation is attempted.

---
## Summary by cubic
Pre-checks memory limits for the `bytes(int)` constructor so large `bytes(n)` raises MemoryError before allocation, matching other repeat/padding ops and preventing host aborts.

- **Bug Fixes**
  - Guard `bytes(n)` with `check_repeat_size(size, 1, vm.heap.tracker())` before allocating the zero-filled buffer.
  - Added tests to reject over-limit `bytes(1_000_000)` and allow `bytes(100)` within limits.

<sup>Written for commit a6b570681647823c31efee125c8c2c90f77a43e3. Summary will update on new commits.</sup>

### DIFF
--- a/crates/monty/src/types/bytes.rs
+++ b/crates/monty/src/types/bytes.rs
@@ -166,6 +166,12 @@ impl Bytes {
                     return Err(ExcType::value_error_negative_bytes_count());
                 }
                 let size = usize::try_from(*n).expect("bytes count validated non-negative");
+                // Pre-check the requested size against resource limits before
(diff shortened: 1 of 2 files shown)


davidhewitt (approved):
Thanks

PR #436, opened 2026-05-14
Pre-check json.dumps indent string size against resource limits

dsp-ant:
The integer indent= argument was converted directly into a repeated space string on the native heap with no resource pre-check, so a large value would allocate well past the configured memory budget. Validate the requested count against the resource tracker before materializing the indent string, raising MemoryError instead. Other .repeat() call sites already carry the same guard.

---
## Summary by cubic
Pre-checks `json.dumps(indent=...)` against resource limits to prevent unbounded indent string allocation. Large values now raise `MemoryError` before allocating, matching guards used at other `.repeat()` sites.

- **Bug Fixes**
  - Validate indent size via `check_repeat_size` before building the space string.
  - Added tests for rejecting huge indents and allowing small indents within limits.

<sup>Written for commit 582ea1b1009534561c5afcab48da2bbcb7ca0936. Summary will update on new commits.</sup>

### DIFF
--- a/crates/monty/src/modules/json/dump.rs
+++ b/crates/monty/src/modules/json/dump.rs
@@ -15,7 +15,7 @@ use crate::{
     exception_private::{ExcType, RunResult},
     heap::{DropWithHeap, HeapData, HeapGuard, HeapId, HeapReadOutput},
     intern::StaticStrings,
-    resource::ResourceTracker,
+    resource::{ResourceTracker, check_repeat_size},
     sorting::{apply_permutation, sort_indices},
     types::{PyTrait, long_int::check_bigint_str_digits_limit, str::allocate_string},
     value::Value,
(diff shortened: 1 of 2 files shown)


davidhewitt (commented):
This doesn't seem like the right defence to me; the indent can be used multiple times over when dumping so a single check doesn't appear to make sense.

I have been playing around with a `StringBuilder` type which checks against memory limits as the string is formed, I think the correct defence for the JSON module in general is probably to switch to that construction rather than apply a defence at this point?

davidhewitt:
I pushed the `StringBuilder` experiment to #448, I will see how it looks here.

samuelcolvin:
@davidhewitt is this now fixed?

davidhewitt:
I have a branch (not pushed yet) which uses `StringBuilder` in `json.dumps`. The challenge is that the `ResourceTracker` has to be slightly decoupled from the `Heap` so that the builder can have a reference to the tracker while the VM is busy traversing heap values. Most likely will finish this week after next.

samuelcolvin:
should be fixed by `StringBuilder` and global allocator.

PR #438, opened 2026-05-14
Enforce a recursion-depth ceiling in json.dumps serialization

dsp-ant:
The serialize_value/serialize_sequence/serialize_dict helpers are mutually recursive and run entirely within one bytecode instruction, so the VM's per-instruction recursion check never fires. The cycle detector catches reference loops but not deep acyclic chains, which can overflow the host's native stack.

Add a depth ceiling matching the JSON_RECURSION_LIMIT already used by json.loads, raising a catchable RecursionError once exceeded.

---
## Summary by cubic
Adds a recursion-depth ceiling to `json.dumps` to prevent native stack overflows on deeply nested acyclic structures. Matches the `json.loads` limit and raises a catchable `RecursionError` when exceeded.

- **Bug Fixes**
  - Enforce `JSON_DUMP_RECURSION_LIMIT` (200) in the serializer and map overflow to `RecursionError`.
  - Add tests: deep nesting triggers the error; moderate nesting stays within the limit.

<sup>Written for commit 058201206c421e63797f643ab70660abc45d6ab9. Summary will update on new commits.</sup>

### DIFF
--- a/crates/monty/src/modules/json/dump.rs
+++ b/crates/monty/src/modules/json/dump.rs
@@ -15,7 +15,7 @@ use crate::{
     exception_private::{ExcType, RunResult},
     heap::{DropWithHeap, HeapData, HeapGuard, HeapId, HeapReadOutput},
     intern::StaticStrings,
-    resource::ResourceTracker,
+    resource::{ResourceError, ResourceTracker},
     sorting::{apply_permutation, sort_indices},
     types::{PyTrait, long_int::check_bigint_str_digits_limit, str::allocate_string},
     value::Value,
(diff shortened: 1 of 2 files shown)


crates/monty/src/modules/json/dump.rs
```diff
@@ -367,6 +367,15 @@ fn json_separator_to_string(value: &Value, role: &str, vm: &VM<'_, impl Resource
...
 }
 
+/// Maximum nesting depth accepted by `json.dumps()`.
+///
+/// Mirrors `JSON_RECURSION_LIMIT` on the `json.loads()` side: serialization is
+/// mutually recursive across `serialize_value`/`serialize_sequence`/`serialize_dict`
+/// and would otherwise overflow the host's native stack on a deep (acyclic)
+/// structure that the cycle detector cannot catch. CPython raises a catchable
+/// `RecursionError` here; we map the depth-limit `ResourceError` to the same.
+const JSON_DUMP_RECURSION_LIMIT: usize = 200;
```
davidhewitt, line 377: Why does this use a constant rather than respect the `ResourceLimit` recursion limit?

Separately, it seems like CPython does not, a 20000-deep nested tuple will serialize fine even if `sys.setrecursionlimit` is set very low.

Seems like CPython uses stack overflow protections for these cases, if I write a million element tuple I eventually get

```
RecursionError: Stack overflow (used 8144 kB) while encoding a JSON object
```

... maybe #440?

davidhewitt:
I decided to solve this case with the general recursion limit mechanism for now, see #454 

Thanks again for the report & patch, will close this one.

PR #439, opened 2026-05-14
Bound except-clause tuple matching by the recursion-depth limit

dsp-ant:
The exception type passed to an except clause is constructed at runtime, so a tuple of types can be nested far deeper than the source-nesting limit allows. The matcher recurses on the host's native stack inside a single bytecode instruction, so deep nesting overflows the host stack rather than raising a Python exception.

Use the heap's recursion-depth tracker (and time check) to bound the traversal, raising RecursionError before the native stack is exhausted.

---
## Summary by cubic
Aligns `except` clause matching with CPython: only classes or flat tuples are valid. Nested tuples now raise `TypeError` (“catching classes that do not inherit from BaseException is not allowed”), removing recursion and preventing host stack overflows; flat-tuple matching is unchanged.

- **Bug Fixes**
  - Reject nested tuples and non-exception items; accept only single classes or flat tuples.
  - Validate the whole tuple (no short-circuit) to match CPython.
  - Route the validity `TypeError` through normal exception handling so it propagates out of the try.
  - Added parity tests: nested tuples raise; empty tuple never matches and lets the original exception propagate; large flat tuple matches; deep runtime nesting raises `TypeError` without overflowing.

<sup>Written for commit 5b2492516a544772746a2f17477dc6806829ee26. Summary will update on new commits. <a href="https://cubic.dev/pr/pydantic/monty/pull/439?utm_source=github">Review in cubic</a></sup>


crates/monty/tests/resource_limits.rs
```diff
@@ -2039,3 +2039,59 @@ len(x) + len(d) + len(s)
...
+    let code = r"
+t = (ValueError,)
+for _ in range(100):
+    t = (t,)
+try:
+    raise TypeError()
+except t:
+    pass
+except TypeError:
+    pass
```
davidhewitt: Seems like we diverge from CPython here:

```
>>> try:
...     raise TypeError()
... except ((ValueError,),):
...     pass
...
Traceback (most recent call last):
  File "<python-input-4>", line 2, in <module>
    raise TypeError()
TypeError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<python-input-4>", line 3, in <module>
    except ((ValueError,),):
        pass
TypeError: catching classes that do not inherit from BaseException is not allowed
```

Apparently CPython doesn't allow recursive descent into sub-tuples in this particular position?

Looks like (at least in this position) we can have a different defence, we can write the test as a regression test in `test_cases` so we can keep parity with CPython going forwards.

We might still need the full recursion defence here depending on where else this function is used.

crates/monty/tests/resource_limits.rs
```diff
@@ -2039,3 +2039,59 @@ len(x) + len(d) + len(s)
...
+}
+
+/// Test that shallow tuple-of-types matching still works under a recursion limit.
+#[test]
+fn except_shallow_tuple_within_recursion_limit() {
+    let code = r"
+ok = False
+try:
+    raise TypeError()
+except ((ValueError,), (KeyError, TypeError)):
```
davidhewitt: CPython rejects this one with the same `TypeError: catching classes that do not inherit from BaseException is not allowed`.

davidhewitt (approved):
Thanks, for this one I think we can apply a different mitigation.

davidhewitt (changes requested):
Sorry that was meant to be request changes, not approve :)

davidhewitt (approved):
Thanks!

PR #458, opened 2026-05-20
fix: name-based identity and equality for external function values

### LINKED ISSUE #345: Python bindings export external-function values inconsistently depending on name interning
## Summary

The Python bindings export the same logical external-function value differently depending on whether the callable's `__name__` was already interned by the compiled Monty source.

This means a tiny source change like mentioning `foo` earlier in the program changes the final Python result for an unrelated input binding `bar`.

## Repro

```python
import pydantic_monty


def run(code: str, *, inputs=None):
    input_names = list(inputs.keys()) if inputs else None
    return pydantic_monty.Monty(code, inputs=input_names).run(inputs=inputs)


def foo():
    pass


print(run("bar; bar", inputs={"foo": foo, "bar": foo}))
print(run("foo; bar", inputs={"foo": foo, "bar": foo}))
```

## Actual behavior

```python
foo
<function 'foo' external>
```

The final value in both cases is logically the same external function value, but it exports differently.

I also checked a few nearby cases:

```python
run("bar", inputs={"foo": foo, "bar": foo})      # 'foo'
run("foo", inputs={"foo": foo, "bar": foo})      # "<function 'foo' external>"
run("bar, bar", inputs={"foo": foo, "bar": foo}) # ('foo', 'foo')
run("foo, bar", inputs={"foo": foo, "bar": foo}) # ("<function 'foo' external>", "<function 'foo' external>")
```

If I rename the Python callable:

```python
foo.__name__ = "baz"
print(run("bar; bar", inputs={"foo": foo, "bar": foo}))
print(run("foo; bar", inputs={"foo": foo, "bar": foo}))
```

both results become bare `'baz'`, which suggests the trigger is the callable's own name being interned, not the input binding name.

## Expected behavior

The same logical external-function value should export consistently regardless of whether its name happened to appear in the source code.

I would expect one of these:

- always export the same placeholder string/repr-like value
- always export a dedicated Python-side wrapper/object
- reject raw external-function final outputs consistently

But it should not depend on whether the function name was interned during compilation.

## Likely cause

Python callables passed through `inputs=` are converted to `MontyObject::Function { name, .. }` using `callable.__name__`.

When that enters the VM, `MontyObject::Function` is lowered in two different ways:

- if `vm.interns.get_string_id_by_name(&name)` succeeds, it becomes inline `Value::ExtFunction(StringId)`
- otherwise it becomes heap `HeapData::ExtFunction(String)`

On the way back out, those two runtime representations are handled differently:

- `HeapData::ExtFunction(String)` is converted back to `MontyObject::Function { name, .. }`, which the Python bindings export as the bare name string
- `Value::ExtFunction(StringId)` does not have a dedicated `MontyObject` arm and falls through to `repr_or_error(...)`, which eventually exports as `"<function 'foo' external>"`

Relevant code:

- `crates/monty-python/src/convert.rs`
- `crates/monty/src/object.rs`

This seems separate from #339. #339 is about dispatch using `callable.__name__`; this issue is about inconsistent final output/export of the resulting external-function value depending on name interning.

### LINKED ISSUE #347: External function identity depends on whether the name appears in source
## Summary

For callable inputs, merely mentioning the callable's name in the Monty source can change `is` behavior for unrelated bindings to the same Python callable.

This is a runtime semantics issue, not just a Python-export formatting issue.

## Repro

```python
import pydantic_monty


def foo():
    pass


inputs = {'a': foo, 'b': foo}


def run(code):
    return pydantic_monty.Monty(code, inputs=list(inputs)).run(inputs=inputs)


assert run('(a is b, a == b)') == (False, False)
assert run('foo = 1; (a is b, a == b)') == (True, False)
```

## Actual behavior

Monty returns:

```python
(False, False)
(True, False)
```

So adding `foo = 1;` changes the identity result for `a is b`, even though `a` and `b` were already provided as the same Python callable input.

## Expected behavior

Mentioning `foo` in the source should not change the identity semantics of `a` and `b`.

At minimum, I would expect the result to be stable across those two snippets.

## Notes

This appears related to external function values having two runtime representations:

- inline `Value::ExtFunction(StringId)` when the callable name is already interned
- heap `HeapData::ExtFunction(String)` otherwise

The source mention seems to affect whether the callable name is interned, which then affects identity behavior.

Relevant code:

- `crates/monty/src/object.rs`
- `crates/monty/src/value.rs`

This seems distinct from #345, which is about final Python export of external-function values depending on name interning. Here the inconsistent behavior is visible inside Monty itself via `is`.

mbeschastn0v:
Fixes #347 (external function identity). Fixes #345 (inline ExtFunction export).

## What changed

`Value::is` and `Value::id` are now name-aware for `ExtFunction` values across both inline (`Value::ExtFunction`) and heap (`HeapData::ExtFunction`) representations. Same-named external functions now satisfy CPython's `a is b ⇒ id(a) == id(b)` invariant regardless of how each side was constructed. Both methods now take `&VM` (shared, read-only — matching the `PyTrait::py_type` precedent) so the heap is reachable for the cross-representation arms.

`Value::py_eq` and `HeapData::py_eq` gained matching cross-representation arms, and ExtFunction hashing is aligned via `hash_python_str` on both sides so the equality/hash contract holds (required for dict keys, set membership, and \`in\` checks).

\`object.rs::from_value_inner\` now has the missing \`Value::ExtFunction → MontyObject::Function\` arm — that was #345 in one line.

## Design note

An earlier iteration of this branch introduced parallel \`Value::py_is\` and \`Value::py_id\` methods alongside the existing identity-only \`is\`/\`id\`. After review discussion, we collapsed that to modify-in-place: \`is\` and \`id\` themselves now carry the name-aware semantics, with the display-only callers in \`py_repr_fmt\` updated to pass \`vm\`. One method per concept, no parallel surface to keep in sync.

## Coverage

Cross-representation \`ExtFunction\` state is reachable via the REPL (interns accumulate across feed_run calls — see \`repl.rs:252\`), and \`tests/external_function_identity.rs::repl_cross_representation_extfunction_identity\` exercises that scenario directly. \`Value::is\` is now a one-liner delegating to \`Value::id\`, which itself handles the cross-representation case, so the explicit per-case arms only remain where the id fallback wouldn't apply (\`Value::py_eq\` and \`HeapReadOutput::py_eq\`).

## Commit shape

History was rewritten via force-push to a single commit (squashed from 6 incremental commits) reflecting the final modify-in-place design. The squash is based on the latest \`origin/main\`.

## Test plan

- [x] \`cargo test -p monty external_function_identity\` (9 Rust integration cases, including REPL cross-rep)
- [x] \`pytest crates/monty-python/tests/test_external_function_identity.py\` (8 cases)
- [x] \`make test-memory-model-checks\`
- [x] Full datatest sweep (946 cases) green
- [x] \`make lint-rs\`, \`make lint-py\` clean
- [x] Existing \`tests/name_lookup.rs::resolve_function_with_non_interned_name\` passes unchanged

### DIFF
--- a/crates/monty/src/builtins/id.rs
+++ b/crates/monty/src/builtins/id.rs
@@ -11,7 +11,7 @@ pub fn builtin_id(vm: &mut VM<'_, impl ResourceTracker>, args: ArgValues) -> Run
     let value = args.get_one_arg("id", vm.heap)?;
     defer_drop!(value, vm);
 
-    let id = value.id();
+    let id = value.id(vm);
 
     // Python's id() returns a signed integer; reinterpret bits for large values
     // On 64-bit: large addresses wrap to negative; on 32-bit: always fits positive
--- a/crates/monty/src/bytecode/vm/compare.rs
+++ b/crates/monty/src/bytecode/vm/compare.rs
@@ -58,14 +58,6 @@ impl<T: ResourceTracker> VM<'_, T> {
     }
 
     /// Identity comparison (is/is not).
-    ///
-    /// Compares identity using `Value::is()` which compares IDs.
-    ///
-    /// Identity is determined by `Value::id()` which uses:
-    /// - Fixed IDs for singletons (None, True, False, Ellipsis)
-    /// - Interned string/bytes index for InternString/InternBytes
-    /// - HeapId for heap-allocated values (Ref)
-    /// - Value-based hashing for immediate types (Int, Float, Function, etc.)
     pub(super) fn compare_is(&mut self, negate: bool) {
         let this = self;
 
@@ -74,7 +66,7 @@ impl<T: ResourceTracker> VM<'_, T> {
         let lhs = this.pop();
         defer_drop!(lhs, this);
 
-        let result = lhs.is(rhs);
+        let result = lhs.is(rhs, this);
         this.push(Value::Bool(if negate { !result } else { result }));
     }
 
--- a/crates/monty/src/value.rs
+++ b/crates/monty/src/value.rs
@@ -18,7 +18,7 @@ use crate::{
     builtins::Builtins,
     bytecode::{CallResult, VM},
     exception_private::{ExcType, RunError, RunResult, SimpleException},
-    hash::HashValue,
+    hash::{HashValue, hash_python_str},
     heap::{ContainsHeap, DropWithHeap, Heap, HeapData, HeapGuard, HeapId, HeapReadOutput},
     intern::{BytesId, FunctionId, Interns, LongIntId, StaticStrings, StringId},
     modules::ModuleFunctions,
@@ -215,6 +215,19 @@ impl PyTrait<'_> for Value {
             // Module functions equality
             (Self::ModuleFunction(mf1), Self::ModuleFunction(mf2)) => Ok(mf1 == mf2),
             (Self::DefFunction(f1), Self::DefFunction(f2)) => Ok(f1 == f2),
+            // External function equality is name-based: two ExtFunction values
+            // (any combination of inline `Value::ExtFunction(StringId)` and heap
+            // `HeapData::ExtFunction(String)`) compare equal iff they reference
+            // the same function name. Interned StringIds are deduped, so equal
+            // StringId ⇔ equal string; the cross-representation arms read the
+            // heap string directly. (#347)
+            (Self::ExtFunction(s1), Self::ExtFunction(s2)) => Ok(s1 == s2),
+            (Self::ExtFunction(s), Self::Ref(id)) if let HeapData::ExtFunction(n) = vm.heap.get(*id) => {
+                Ok(interns.get_str(*s) == n.as_str())
+            }
+            (Self::Ref(id), Self::ExtFunction(s)) if let HeapData::ExtFunction(n) = vm.heap.get(*id) => {
+                Ok(n.as_str() == interns.get_str(*s))
+            }
             // Markers compare equal if they're the same variant
             (Self::Marker(m1), Self::Marker(m2)) => Ok(m1 == m2),
             // Properties compare equal if they're the same variant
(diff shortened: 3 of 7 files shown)


mbeschastn0v:
Closing to re-trigger CI — `recursion__deep_hash.py` timed out under `cargo llvm-cov` instrumentation (2s budget), reproduced locally at 0.65-0.93s across 3 runs. Suspected CI runner load flake; no code path in this PR connects to deep tuple/frozenset hash performance.

uv.lock
samuelcolvin, line 1: this change looks unrelated.

crates/monty/src/value.rs
```diff
@@ -1475,11 +1499,28 @@ impl Value {
...
     ///
-    /// Compares value identity by comparing their IDs.
-    pub fn is(&self, other: &Self) -> bool {
-        self.id() == other.id()
+    /// `ExtFunction` values (inline `Value::ExtFunction` or heap
+    /// `HeapData::ExtFunction`) compare identical when their function names
+    /// match — the closest approximation Monty can offer after
+    /// `MontyObject::Function` conversion has discarded host object identity.
+    /// All other variants compare by [`Value::id`].
+    pub fn is(&self, other: &Self, vm: &VM<'_, impl ResourceTracker>) -> bool {
```
samuelcolvin, line 1505: do we still need this change?

I feel like the previous `self.id() == other.id()` implementation should still work for all cases, albeit with slightly worse performance than these tweaks.

crates/monty/src/bytecode/vm/compare.rs
```diff
@@ -59,9 +59,10 @@ impl<T: ResourceTracker> VM<'_, T> {
 
     /// Identity comparison (is/is not).
     ///
-    /// Compares identity using `Value::is()` which compares IDs.
-    ///
-    /// Identity is determined by `Value::id()` which uses:
+    /// Dispatched through [`Value::is`]. For `ExtFunction` values the
```
samuelcolvin: this extra comment verbosity seems unnecessary here.

crates/monty/src/builtins/id.rs
```diff
@@ -7,11 +7,13 @@ use crate::{
 /// Implementation of the id() builtin function.
 ///
 /// Returns the identity of an object (unique integer for the object's lifetime).
+/// `ExtFunction` values get a name-derived id via [`Value::id`], matching the
```
samuelcolvin: again, I don't think we need the extra lines in the docstring here.

samuelcolvin (commented):
overall, I think this looks good. just a few things to tweak

uv.lock
mbeschastn0v: Reverted in feb55df — uv.lock no longer in the diff. Heads up: my local uv 0.11.15 strips the `[options]` block on every resolve ("Resolving despite existing lockfile due to removal of global exclude newer"), so I had to `SKIP=lint-rs` for the amend to keep the revert. CI runs its own uv version that doesn't strip, so the lockfile stays clean in the PR.

crates/monty/src/value.rs
mbeschastn0v: Good catch — collapsed to `self.id(vm) == other.id(vm)` in feb55df. Since `Value::id` is already name-aware for `ExtFunction` (any representation), the explicit match arms here were redundant. Saves ~15 lines and closes most of the codecov gap.

crates/monty/src/bytecode/vm/compare.rs
mbeschastn0v: Trimmed to a one-liner in feb55df.

crates/monty/src/builtins/id.rs
mbeschastn0v: Reverted to the original one-line docstring in feb55df.

mbeschastn0v:
Thanks for your review @samuelcolvin! I have pushed the changes, please have a look

samuelcolvin:
Great. Thank you!

We should update [limitations](https://github.com/pydantic/monty/tree/main/limitations) to reflect these changes, I'll do that now.

PR #467, opened 2026-05-26
proc macro for function signature parsing

samuelcolvin:
fix #294.

TODO:
* [x] use `FromArgs` in `OsFunction`s

---
## Summary by cubic
Introduces `monty-macros` proc‑macros and converts function/method/OS‑call parsing to typed structs with CPython‑accurate errors. Replaces raw `OsFunction` dispatch with a typed `OsFunctionCall`, updates REPL/async paths, and unifies `sorted()`/`list.sort` kwargs and duplicate‑kwarg errors on methods.

- **New Features**
  - New crate `monty-macros` with `#[derive(FromArgs)]` and `#[derive(ToArgs)]`; core traits `FromValue`/`ToMontyObject`.
  - End‑to‑end typed `OsFunctionCall`: `.to_args()`, `.name()`, `.on_no_handler()`, move‑by‑value via `take_function_call()` and a `Used` placeholder; mount table now handles `&OsFunctionCall`.
  - `MethodDictMerge` opcode qualifies duplicate‑kwarg errors with the receiver type; `sorted()` delegates to `parse_and_sort` for shared kwargs/errors.
  - `LaxBool` for Python truthiness in kwargs (e.g., `list.sort(reverse=...)`, `Path.mkdir(parents=..., exist_ok=...)`).
  - Macro adoption across builtins/modules: `open`, `str.encode`, `bytes.decode`, `str`/`bytes` split/replace/expandtabs, `map`, `min/max`, `pow`, `print`, `zip`, `re.sub/split` and `Pattern.sub/split`, `dict`/`update`, `re.Match.groupdict`, `date`/`datetime`/`timezone`/`timedelta`, `json.dumps`, `list.sort`.

- **Bug Fixes**
  - CPython parity for argument errors: `_PyArg_BadArgument` via `#[from_args(bad_arg|bad_arg_named)]`; exact arity wording (`expected_exact`, singular/plural), deferred unknown‑kwarg ordering, and “keyword argument” conflict text; method vs function names match CPython.
  - `math.isclose` accepts `a`/`b` kwargs and validates tolerances; `map()` emits the CPython “at least two arguments” message; `json.dumps` arity matches C‑style wording.
  - `open(mode=...)`, `encode`/`decode` produce named bad‑arg errors; `Path.mkdir(mode=0o777, parents=False, exist_ok=False)` has the correct signature and truthiness; `timezone(td, None)` now raises `TypeError` and reports `None` (not `NoneType`) in the bad‑arg message.
  - Removed brittle helpers (`get_zero_one_named_arg`, `get_zero_one_two_args`, `extract_keyword_only_pair`) in favor of the macro for consistent errors and refcount safety.

<sup>Written for commit c2bca74d7ecb38422995bbb2bc7071b8fe35e860. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/467?utm_source=github">Review in cubic</a>


crates/monty-macros/src/from_args.rs
```diff
@@ -0,0 +1,1358 @@
...
+    /// Fields in declaration order — also the order of positional arguments.
+    fields: Vec<Field>,
+    /// Index of the `*args` field (if any).
+    varargs_idx: Option<usize>,
+    /// Index of the `**kwargs` field (if any).
+    varkwargs_idx: Option<usize>,
+    /// Which `type_error_c_at_most*` helper to emit when too many positional
+    /// arguments are passed. Only meaningful in [`ErrorStyle::C`] — the named
+    /// styles always emit the `{name}() takes at most …` helper.
+    at_most_style: AtMostStyle,
```
davidhewitt: Perhaps move this as a field inside `ErrorStyle::C`?

crates/monty-macros/src/from_args.rs
```diff
@@ -0,0 +1,1358 @@
...
+    /// exclusive with `varargs`/`varkwargs` — the pre-count is meaningless
+    /// when the signature accepts unbounded inputs.
+    at_most_total: bool,
+    /// When set, the generated code pre-validates that *at least* the
+    /// required-positional count appears as positional args (kwargs do
+    /// not fill the quota). Emits CPython's method-style wording:
+    /// `{name}() takes at least M positional argument(s) (N given)` —
+    /// matches `str.replace` and similar C-implemented methods that
+    /// reject kwargs for the required positionals.
+    at_least_positional: bool,
```
davidhewitt: Seems like this is probably just required positional-only arguments and should be a property derived from the function shape rather than a bool here.

crates/monty-macros/src/from_args.rs
```diff
@@ -0,0 +1,1358 @@
...
+    /// positional count equals the maximum. Mutually exclusive with
+    /// `varargs`, `at_most_total`, and `at_least_positional`.
+    expected_exact: bool,
+    /// Optional override for the function name used in the
+    /// unknown-kwarg error (`{name}() got an unexpected keyword argument 'X'`).
+    /// Used by `sorted()` because CPython's sorted() delegates internally
+    /// to `list.sort` and surfaces sort()'s kwarg error wording, so the
+    /// kwarg-name error has to read `sort()` while arity errors keep
+    /// using the struct's primary `name = "sorted"`.
+    kwarg_error_name: Option<String>,
```
davidhewitt: I wonder if the better solution here is to remove this special case from the arg parsing code and instead make the `sorted` builtin do the same thing where it creates a new list from the iterable and then calls `.sort()` on it.

crates/monty-macros/src/from_args.rs
```diff
@@ -0,0 +1,1358 @@
...
+    /// including the special `None` rendering for `NoneType` values via
+    /// [`Type::cpython_arg_name`]. Replaces the inner `FromValue` error
+    /// (e.g. the generic `"a str is required"`) so migrating a C-extension
+    /// function to the macro keeps the original wording. The variant
+    /// chooses between CPython's positional (`argument N`) and named
+    /// (`argument 'X'`) phrasings; CPython picks one or the other per
+    /// function (`strftime` uses positional, `open`/`encode`/`decode` use
+    /// named). Only fires when `EXPECTED_TYPE_NAME` is `Some(_)`; the
+    /// identity `Value` impl falls through to its native error.
+    bad_arg: Option<BadArgStyle>,
```
davidhewitt, line 46: FWIW, it looks like the difference might be that `PyArg_ParseTupleAndKeywords` uses index-based, whereas argument clinic uses name-based.

It looks like `strftime` will use also use name-based on Python 3.15 because it got converted to use argument clinic.

I wonder if some of these options are therefore coupled, e.g. `at_most_total` mentions `PyArg_ParseTupleAndKeywords` too.

crates/monty-macros/src/from_args.rs
```diff
@@ -0,0 +1,1358 @@
...
+                    // by name without needing to thread them through helpers.
+                    macro_rules! __cleanup {
+                        ($err:expr) => {{
+                            #cleanup_block
+                            // Also drop anything left in the iterators.
+                            __pos_iter.drop_with_heap(heap);
+                            __kwargs_iter.drop_with_heap(heap);
+                            return Err($err);
+                        }};
+                    }
```
davidhewitt: I think rather than this approach, I suggest the slots have a custom struct with a `DropWithHeap` implementation and a `.finish()` method to build the final struct.

Then various `HeapGuard` / `defer_drop` / `defer_drop_mut!` handling should make it possible to guarantee cleanup on error exit without needing to use a macro to duplicate cleanup into every exit branch.

Will lead to significantly leaner generated code, more efficient & faster compile :)

crates/monty-macros/src/from_args.rs
```diff
@@ -0,0 +1,1358 @@
...
+
+                    #total_check
+                    #exact_check
+                    #at_least_check
+                    #unknown_decl
+                    #positional_loop
+                    #kwarg_loop
+                    #missing_check
+                    #unknown_check
```
davidhewitt, line 364: It looks like most of these are fully inlined code generated by the macros. I'd expect that some of the pieces (e.g. the checks, maybe the loops) could have helper functions extracted into the `monty` crate. This would probably enable the optimizer to do a better job / faster compile times / smaller binary size because there'd be less overall stuff to chew on.

crates/monty-macros/README.md
davidhewitt, line 1: I think a lot of this stuff doesn't want to be in the `README` and instead could be in the docs on the macro attributes themselves?

crates/monty-python/src/async_dispatch.rs
```diff
@@ -334,22 +334,22 @@ fn dispatch_function_call(
...
-                Some(format!("OS function '{function}' not implemented")),
+                Some(format!("OS function '{}' not implemented", call.name())),
             )
             .into();
         };
 
+        // Project the typed args back to the host-facing
+        // `(positional, keyword)` shape. Cloning is fine — `OsFunctionCall`
+        // is owned by the OsCall struct and we just want to read it.
+        let (args, kwargs) = call.clone().to_args();
```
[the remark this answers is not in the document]
davidhewitt: See comment just above.

crates/monty/src/args/mod.rs
```diff
@@ -1,18 +1,34 @@
...
 
+/// Projects a typed args struct into the `(positional, keyword)` `MontyObject`
+/// pair host callbacks expect. Consumes `self` to avoid cloning owned fields.
+///
+/// Inverse of [`FromArgs`]: `FromArgs` is internal `ArgValues` → struct,
+/// `ToArgs` is struct → host-facing `(args, kwargs)`. Driven by
+/// [`crate::os::OsFunctionCall::to_args`] for the monty-python / monty-js bindings.
+pub trait ToArgs {
+    fn to_args(self) -> (Vec<MontyObject>, Vec<(MontyObject, MontyObject)>);
+}
```
davidhewitt, line 30: This is _not_ the inverse of `FromArgs`, as it's about the host-side.

If we want this to be the inverse it should go to `ArgValues`, and then let callers do `ArgValues::into_py_objects`?

crates/monty/src/modules/asyncio.rs
```diff
@@ -99,12 +99,12 @@ fn run(heap: &mut Heap<impl ResourceTracker>, args: ArgValues) -> RunResult<Call
 /// # Errors
 /// Returns `TypeError` if any argument is not awaitable.
 pub(crate) fn gather(heap: &mut Heap<impl ResourceTracker>, args: ArgValues) -> RunResult<Value> {
-    let (pos_args, kwargs) = args.into_parts();
+    // TODO: support keyword arguments (e.g. return_exceptions); for now any
+    // kwarg is rejected by `into_pos_only` with the standard
+    // "gather() does not support keyword arguments" TypeError.
+    let pos_args = args.into_pos_only("gather", heap)?;
     defer_drop_mut!(pos_args, heap);
```
davidhewitt: Potentially can make `kwargs_not_supported_yet` an option on the macro so that these kinds of callsites can also be cleaned up.

PR #473, opened 2026-05-28
fix: accept subscript and attribute targets in unpacking, for-loop, and comprehension contexts

### LINKED ISSUE #408: Subscript targets fail in unpacking assignment
Monty raises `MontySyntaxError: invalid unpacking target` for valid Python unpacking assignments when the assignment target is a subscript.

This came up in agent-generated code that tried to swap two list elements:

```python
x[i], x[j] = x[j], x[i]
```

That is valid Python, but Monty rejects the subscript assignment target during unpacking.

### Minimal reproduction

```python
from pydantic_monty import MontyRepl

code = "x=[0]\nx[0], = (1,)\nx"
result = MontyRepl(script_name="repro.py").feed_run(code)
print(result)
```

### Expected behavior

This should match CPython and print:

```python
[1]
```

### Actual behavior

Monty raises:

```text
MontySyntaxError: invalid unpacking target: Subscript(ExprSubscript { node_index: NodeIndex(None), range: 6..10, value: Name(ExprName { node_index: NodeIndex(None), range: 6..7, id: Name("x"), ctx: Load }), slice: NumberLiteral(ExprNumberLiteral { node_index: NodeIndex(None), range: 8..9, value: Int(0) }), ctx: Store })
```

### Additional examples

These also fail:

```python
x = [1, 2]
x[0], x[1] = x[1], x[0]
```

```python
x = [1, 2]
i, j = 0, 1
x[i], x[j] = x[j], x[i]
```

But these work:

```python
x = [1, 2]
x[0] = 3
```

```python
i, j = 0, 1
i, j = j, i
```

So the issue appears specific to unpacking assignment targets that are subscripts, not subscript assignment in general.

### Environment

Reproduced with:

- `pydantic-monty==0.0.16`
- `pydantic-monty==0.0.17`
- Python 3.13.0
- macOS arm64

---

By the way, huge thanks again for building Monty! Let me know if these kinds of reports are useful and if should keep filing them based on our observations of using Monty. Or if it's too soon in Monty's development stage.

mbeschastn0v:
Closes #408.

## Summary

CPython allows subscript (`x[i]`) and attribute (`o.a`) expressions as targets inside tuple unpacking, `for`-loop variables, and comprehension `for` clauses. Monty's `parse_unpack_target_impl` only accepted `Name`/`Tuple`/`Starred`/`List` and rejected everything else with `SyntaxError: invalid unpacking target: <kind>`. That breaks:

- `x[0], x[1] = x[1], x[0]` (the reported case)
- `x[0], = (1,)`
- `o.a, o.b = 1, 2` (attribute targets — same parse arm, wider than the issue title)
- `for box[0] in iter:` / `for o.a in iter:`
- `[... for box[0] in iter]` and the same shape in comprehensions

This PR fixes all five contexts in a single parse-site change plus the matching prepare/compile wiring.

## Design

`UnpackTarget` gains two new leaf variants `Subscript { target, index, target_position }` and `Attribute { object, attr, target_position }`, mirroring `AssignTarget::Subscript`/`Attr` exactly. The two-enum split between `AssignTarget` (top-level assignment LHS) and `UnpackTarget` (inside-an-unpack pattern) is intentional grammar typing — `Starred` is only valid inside unpack, `Unpack` is only valid at the top — and the fix preserves that. The leaves themselves are duplicated between the two enums (matching how `Name` is duplicated today); a follow-up could extract a shared `LeafAssignTarget` if a third enum ever needs the same leaves, but YAGNI for now.

## Implementation notes

- **Compiler stack invariant**: `compile_unpack_target` and `process_unpack_sim` delegate to `emit_subscript_store`/`emit_attr_store`. Each variant's net operand-stack effect is −1, identical to `Name`/`Starred`, so the existing per-target loop and `compile_comp_target_unpack`'s leaf enumeration work unchanged. In the comprehension simulator the sim ↔ operand-stack 1:1 invariant is preserved by shrinking sim in lock-step (the `Pending` is popped and nothing pushed back — no sentinel needed). Traced through `Tuple { sub_0=Subscript, sub_1, sub_2 }` cases to confirm.

- **CPython evaluation order**: RHS unpacks first, then each target's sub-expressions evaluate at store time. Locked in by the `side-effecting index` test in `unpack__ops.py`.

- **Silent-correctness fix**: `collect_cell_vars_from_node` and `collect_referenced_names_from_node` previously ignored `Node::For`/`Node::With` targets — correct then, since `UnpackTarget` had no sub-expressions. After this change the targets can carry expressions that reference enclosing variables, so two new helpers `collect_{cell_vars,referenced_names}_from_unpack_target` are wired into both `For` and `With` arms.

- **Comp-var slot allocation**: `prepare_unpack_target_for_comprehension`'s new arms intentionally do NOT allocate comp-var slots. Subscript/attribute targets mutate existing state, they don't introduce a binding — their sub-expressions resolve against the surrounding scope via the normal `prepare_expression` path.

- **Walrus inside subscript index**: a walrus `(i := 0)` inside the index expression of a subscript target correctly binds in the outer scope — `collect_names_from_unpack_target` walks the sub-expressions via `collect_assigned_names_from_expr` to pick it up. Locked in by the `walrus in subscript index` test.

## Test plan

- [x] `cargo test -p monty --features memory-model-checks` (all rust tests pass, 968 cases)
- [x] `cargo test -p monty --test parse_errors --features memory-model-checks` (63 cases, including new negative snapshots)
- [x] `cargo run -p monty-datatest --features memory-model-checks unpack__` / `iter__for_loop` / `comprehension__` (all green)
- [x] `make test-py` (1091 Python binding tests pass)
- [x] CPython parity confirmed for every new test case
- [x] `make format-rs && make lint-rs && make lint-py` clean
- [x] Pre-commit hooks pass

## Tests added

| File | Coverage |
|---|---|
| `crates/monty/test_cases/unpack__ops.py` | subscript-target assignment unpack: single, two, swap, nested in tuple, name siblings, starred middle, walrus in index, side-effecting index |
| `crates/monty/test_cases/iter__for_loop_unpacking.py` | `for x[i] in iter:` last-value, computed index re-evaluated each iteration, body visibility, nested tuple |
| `crates/monty/test_cases/comprehension__all.py` | comp `for` clause with subscript targets, nested tuple subscript, leak-out semantics |
| `crates/monty/test_cases/unpack__attr.py` (new) | attribute targets across assignment / swap / nested / starred / for-loop / comprehension via `make_mutable_point()` |
| `crates/monty/tests/parse_errors.rs` | replaced `for_loop_attribute_target_has_clean_message` (the snapshot that locked the now-fixed rejection) with `unpack_target_constant_rejected` / `unpack_target_call_rejected` to lock the upstream ruff rejection for genuinely-invalid LHS shapes |

## Out of scope

- Refactoring the leaf overlap between `AssignTarget` and `UnpackTarget` into a shared `LeafAssignTarget` — preserved the existing architecture; happy to split that out if reviewers prefer.
- Augmented assignment (`x[0] += 1`) — separate code path via `AugAssign`, unchanged.

---
## Summary by cubic
Allow subscript (`x[i]`) and attribute (`obj.a`) targets in unpacking, for-loops, and comprehensions, matching CPython. Fixes #408 and unblocks swaps, nested patterns, and comprehension targets.

- **Bug Fixes**
  - Parser/AST: Add `UnpackTarget::Subscript` and `UnpackTarget::Attribute` (with `target_position`); parse arms mirror assign-target shapes.
  - Compiler: Delegate stores to `emit_subscript_store`/`emit_attr_store`; net stack effect −1; preserve CPython eval order; comprehension sim shrinks in lock-step.
  - Prepare/Scope: Recurse into sub-expressions; no comp-var slots for these leaves; walk `For`/`With` targets for cell-var and referenced-name analysis; walrus inside indices binds in outer scope.
  - Tests: Cover assignment, for-loops, and comprehensions (subs/attrs: swaps, nested, starred, side effects, leak-out); add lambda-in-subscript-index cases to exercise closure capture in `for` targets and nested tuple targets; update parse-error snapshots to lock upstream rejections for constant/call LHS shapes.

<sup>Written for commit 51a2d41042cbe20bfbeff1950f54ed590b53867a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/473?utm_source=github">Review in cubic</a>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 /TODO.md
 /monty
 /.claude/settings.local.json
+/CLAUDE.local.md
 /scratch/
 /worktrees/
 /flame/
--- a/crates/monty/src/bytecode/compiler.rs
+++ b/crates/monty/src/bytecode/compiler.rs
@@ -118,12 +118,15 @@ fn check_comp_generators(count: usize, position: CodeRange) -> Result<(), Compil
 /// Returns a position that locates `target` in source for error reporting.
 ///
 /// `Name` / `Starred` carry the identifier's position; `Tuple` carries its
-/// own. Used by comp-target unpacking when the per-leaf position isn't
-/// available at the error point.
+/// own; `Subscript`/`Attribute` carry `target_position`. Used by comp-target
+/// unpacking when the per-leaf position isn't available at the error point.
 fn target_position(target: &UnpackTarget) -> CodeRange {
     match target {
         UnpackTarget::Name(ident) | UnpackTarget::Starred(ident) => ident.position,
         UnpackTarget::Tuple { position, .. } => *position,
+        UnpackTarget::Subscript { target_position, .. } | UnpackTarget::Attribute { target_position, .. } => {
+            *target_position
+        }
     }
 }
 
@@ -2880,6 +2883,22 @@ impl<'a> Compiler<'a> {
             UnpackTarget::Name(ident) | UnpackTarget::Starred(ident) => {
                 sim.push(SimItem::Leaf(ident.namespace_id().as_u16()));
             }
+            UnpackTarget::Subscript {
+                target,
+                index,
+                target_position,
+            } => {
+                // Consumes value off TOS; net −1 matches `Name`/`Starred`, so the
+                // sim ↔ operand-stack 1:1 invariant survives without pushing back.
+                self.emit_subscript_store(target, index, *target_position)?;
+            }
+            UnpackTarget::Attribute {
+                object,
+                attr,
+                target_position,
+            } => {
+                self.emit_attr_store(object, attr, *target_position)?;
+            }
             UnpackTarget::Tuple { targets, position } => {
                 // Pick UNPACK_EX vs UNPACK_SEQUENCE based on whether a starred
                 // sub-target is present (same logic as the regular assignment
@@ -2928,11 +2947,12 @@ impl<'a> Compiler<'a> {
         Ok(())
     }
 
-    /// Compiles storage of an unpack target - either a single identifier, nested tuple, or starred.
+    /// Compiles storage of an unpack target, assuming the value is on TOS.
     ///
-    /// For single identifiers: emits a simple store.
-    /// For nested tuples: emits `UnpackSequence` (or `UnpackEx` with starred) and recursively
-    /// handles each sub-target.
+    /// Each variant's net stack effect is −1, so the per-target loops in
+    /// `Tuple` and `emit_unpack_store` work uniformly. Subscript/attribute
+    /// delegate to `emit_subscript_store`/`emit_attr_store` (sub-expressions
+    /// evaluated at store time, matching CPython).
     fn compile_unpack_target(&mut self, target: &UnpackTarget) -> Result<(), CompileError> {
         match target {
             UnpackTarget::Name(ident) => {
@@ -2943,6 +2963,20 @@ impl<'a> Compiler<'a> {
                 // Starred target by itself (shouldn't happen at top level normally)
                 // Just store as if it were a name
                 self.compile_store(ident)?;
+            }
+            UnpackTarget::Subscript {
+                target,
+                index,
+                target_position,
+            } => {
+                self.emit_subscript_store(target, index, *target_position)?;
+            }
+            UnpackTarget::Attribute {
+                object,
+                attr,
+                target_position,
+            } => {
+                self.emit_attr_store(object, attr, *target_position)?;
             }
             UnpackTarget::Tuple { targets, position } => {
                 // Check if there's a starred target
--- a/crates/monty/src/expressions.rs
+++ b/crates/monty/src/expressions.rs
@@ -357,26 +357,35 @@ pub enum Expr {
     },
 }
 
-/// Target for tuple unpacking - can be a single name, nested tuple, or starred target.
+/// Target for tuple unpacking. Used in assignment LHS, `for`-loop, and
+/// comprehension `for` clauses.
 ///
-/// Supports recursive structures like `(a, b), c` or `a, (b, c)`.
-/// Also supports starred targets like `first, *rest = [1, 2, 3, 4]`.
-/// Used in assignment statements, for loop targets, and comprehension targets.
+/// `Name`/`Starred` introduce bindings; `Subscript`/`Attribute` mutate existing
+/// state and bind nothing (scope analysis must skip them as binding sites but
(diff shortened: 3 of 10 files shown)


mbeschastn0v:
Thanks @cubic-dev-ai — valid catch. The `for a[i] in (10, 20)` test with `i = 0` never reassigned `i` inside the body, so `a == [20, 0]` was satisfied by both the correct re-evaluation behavior and the bad cache-the-index alternative.

Fixed in 3de7147 by mutating `i` inside the loop body so the second store relies on re-evaluation. The discriminating outcome is now `a == [10, 20]` (vs. `[20, 0]` if the index were cached at loop entry).

The comprehension counterpart in `comprehension__all.py` had the same non-discriminating shape but cannot be made discriminating without a body to mutate `i` in — I rephrased its assertion message to drop the "re-evaluated" claim and just state what it actually verifies (subscript store on each emitted value).

Force-pushed: also trimmed verbose comments/docstrings throughout the patch per separate maintainer feedback.

mbeschastn0v:
@samuelcolvin 
Architectural question for follow-up consideration:
Before this PR, AssignTarget and UnpackTarget shared one variant (Name). With the two new leaves this PR adds, they now share three: Name, Subscript, Attribute. That tips the balance toward extracting a shared leaf type, e.g.:
 
```rust
pub enum LeafAssignTarget {
      Name(Identifier),
      Subscript { target: ExprLoc, index: ExprLoc, target_position: CodeRange },
      Attribute { object: ExprLoc, attr: EitherStr, target_position: CodeRange },
 }
 
pub enum AssignTarget {
      Leaf(LeafAssignTarget),
      Unpack { targets: Vec<UnpackTarget>, targets_position: CodeRange },
}
 
pub enum UnpackTarget {
      Leaf(LeafAssignTarget),
      Tuple { targets: Vec<Self>, position: CodeRange },
      Starred(Identifier),
}
```
 
 Wins: store-dispatch centralizes (one emit_leaf_store instead of three parallel match arms in compile_assign_target / compile_unpack_target / process_unpack_sim); collector helpers shrink similarly. The grammar split you encoded — Starred exclusive to unpack, Unpack exclusive to top-level assign — stays intact.
 Cost: every leaf match becomes two-level (UnpackTarget::Leaf(LeafAssignTarget::Subscript { … })), which is patterns getting noisier even when total LoC drops.
 I left it out of this PR to keep the diff scoped to the bug fix, but happy to either bundle it in here or open a follow-up — whichever you prefer. Or skip entirely if you'd rather keep the flat enums; the duplication is real but not large.

PR #481, opened 2026-06-01
Add JS OS call and REPL pause support

ex3ndr:
Hello! JS support for Monty is very limited at the Monty - REPL forces to not use external functions at all and datetime.now functions are all missing which triggers models to struggle a lot. This PR fixes that.


ex3ndr:
It seems to be implemented already

PR #482, opened 2026-06-01
[codex] Add JS virtual filesystem mounts

ex3ndr:
## Summary

Adds first-class JavaScript virtual filesystem mounts for Monty, independent of the external-function support path.

- adds a native paused OS-call surface so unsupported filesystem calls can be handled by JS and resumed
- keeps native `MountDir` handling as the fast path and falls back to paused OS calls only when needed
- adds `VirtualMount` support in the JS wrapper with sync and async backends, read-only/read-write modes, write limits, and path routing
- extends REPL execution with `feedStart`, sync `feed`, and async `feedAsync` support for paused filesystem calls
- adds marker conversions for Monty path, file handle, and stat result values returned from JS

## Validation

- `cargo fmt`
- `cargo check -p monty-js`
- `npm run build:debug`
- `npm test -- __test__/virtual_mount.spec.ts`
- `npm test -- __test__/mount.spec.ts __test__/start.spec.ts __test__/repl.spec.ts __test__/async.spec.ts __test__/virtual_mount.spec.ts`
- `npm run lint`
- `cargo test -p monty-js`
- `git diff --check`
- `npm test`

### DIFF
--- a/crates/monty-js/__test__/virtual_mount.spec.ts
+++ b/crates/monty-js/__test__/virtual_mount.spec.ts
@@ -1,0 +1,168 @@
+import test from 'ava'
+
+import { Monty, MontyRepl, MontyRuntimeError, VirtualMount, VirtualMountError, runMontyAsync } from '../wrapper'
+
+function createMemoryMount(asyncBackend = false): VirtualMount {
+  const files = new Map<string, Buffer>([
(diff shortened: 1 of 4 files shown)


samuelcolvin:
looks like you need to fix conflicts and also address the review-bot comments.

ex3ndr:
It seems you guys almost build this yourself already?

PR #485, opened 2026-06-03
fix(file): release the buffered-file OS-call pin on all paths

arkuhn:
AI disclaimer - this bug was found and fixed with claude code. The changes seem sensible to me, but I'm far from an expert. If the review or verification burden for this change is high, feel free to disregard / I'll happily close.

## What

Fixes two coupled reference-count leaks in the buffered file (`open()`) path.

1. **Pin double-count (read + write).** `inc_ref_for_pending_oscall` inc_ref'd the OpenFile twice, on the assumption that one ref rides in the OS-call args and is dropped at the host boundary. But the buffered OS calls (`OsFunctionCall::ReadText`/`ReadBytes`/`WriteText`/…) carry only a `MontyPath`, never a `Value::Ref` to the file, so that ref is never released. Every buffered `read()`/`write()` leaked one ref on the OpenFile, so the file object was never freed (even with an explicit `close()`). Fixed by pinning with a single `inc_ref` so each release site — the success pin in `apply_buffer_store`/`apply_write_position`, the `dec_ref` in `resume_with_exception`, `VM::drop`, and `CallResult`'s drop — balances exactly one ref.

2. **Buffer leak on free.** `py_dec_ref_ids_for_data` had no `HeapData::OpenFile` arm, so a file read but never `close()`d leaked its full-file buffer (`Str`/`Bytes`) when the OpenFile was freed. Added the arm, plus the matching `for_each_child_id` arm which the in-file comment requires to stay in sync.

## Tests

`crates/monty/tests/file_resource_tracking.rs::dropping_unclosed_file_releases_buffer`: a file read but not closed releases its buffer once the OpenFile is freed (`current_memory()` returns to baseline). This requires both fixes — the OpenFile must be freed (1) and the free path must release the buffer (2).

The pin fix lives in the shared `inc_ref_for_pending_oscall` helper, so it applies identically to writes (same helper + the symmetric single-pin release in `apply_write_position`). The write/append paths in `file_error_paths` and the snapshot/resume round-trips in `binary_serde` pass under `memory-model-checks` (no over-release on either path). A buffered write loads no buffer, so its leak was the small `OpenFile` handle alone, which isn't cleanly isolatable via `current_memory()` — hence the regression test targets the read path, where the leaked buffer makes both fixes observable.

## How this was found

Surfaced while auditing Monty's reference-counting discipline for use-after-free conditions as part of Hack Monty. This one is an over-retention *leak* (memory-safe, not an escape), caught and confirmed with the repo's own `memory-model-checks` + `ref-count-return` tooling.

### DIFF
--- a/crates/monty/src/bytecode/vm/call.rs
+++ b/crates/monty/src/bytecode/vm/call.rs
@@ -88,9 +88,10 @@ impl DropWithHeap for CallResult {
             Self::FramePushed => {}
             Self::OsCallStoreBuffer { call, file_id } => {
                 call.drop_with_heap(heap);
-                let heap = heap.heap_mut();
-                heap.dec_ref(file_id);
-                heap.dec_ref(file_id);
+                // Single pin (see `inc_ref_for_pending_oscall`): release one ref
+                // if the call is discarded before dispatch routes it to a
+                // `pending_file_effect`.
+                heap.heap_mut().dec_ref(file_id);
             }
         }
     }
--- a/crates/monty/src/heap.rs
+++ b/crates/monty/src/heap.rs
@@ -1654,6 +1654,15 @@ fn for_each_child_id<F: FnMut(HeapId)>(data: &HeapData, mut on_child: F) {
                 on_child(tz_id);
             }
         }
+        HeapData::OpenFile(file) => {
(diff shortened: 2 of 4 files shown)


davidhewitt (approved):
Thanks

PR #500, opened 2026-06-11
Move execution to a subprocess pool

samuelcolvin:
## Summary by cubic
Runs Monty in crash‑isolated worker subprocesses via a protobuf protocol and a shared Rust pool for Python and Node.js, improving resilience to crashes/timeouts and tightening wire validation. Also standardizes invalid REPL input errors, eliminates intermediate decode buffers for hot-path payloads to cut peak memory, guards against wire memory amplification, and renames wire messages for clarity without breaking compatibility.

- **New Features**
  - `monty-proto`: protobuf schema with checked‑in `prost` output (exact `prost` pin for reproducible codegen), 4‑byte LE framing, and a hand‑written zero‑copy `MontyObject` wire codec; differential tests ensure byte‑compatibility and validation; cyclic values round‑trip; precise depth checks mirroring prost recursion; MAX_FRAME_LEN enforcement and memory‑amplification guards; no hello handshake; stateless `write_frame`; wire renames for clarity (e.g. `MontyValue` → `MontyObject`, cycle field `heap_id` → `identity`) and broader message cleanups dropping the `*Value` suffixes (`ValueList` → `ObjectList`, `ExceptionValue` → `Exception`, raised exceptions as `RaisedException`) — all wire‑compatible; named‑tuple/dataclass bodies and function/OS‑call args/kwargs decode/encode directly with no intermediate buffers, reducing peak memory and overhead on the hottest paths, with new differential cases.
  - `monty --subprocess`: child mode with strict turn‑based protocol, streamed prints, child‑local mounts, OS callback bubbling, dump/load (preserves script name and type‑check state), clean fatal errors, and an empty worker environment by default.
  - `monty-pool`: elastic worker pool with prewarming, crash detection/replacement, recycling, and a watchdog enforcing per‑request timeouts; improved protocol‑violation classification, deadline handling, and guaranteed child reaping.
  - Python: pools and sessions (`checkout(...).feed_run(...)`), coroutine callbacks, streamed prints, `MontyCrashedError`; `pydantic-monty-cli` installs the `monty` binary; `_binary.py` resolves it; fixes session lifecycle deadlock and execution‑duration accounting.
  - Node.js: `@pydantic/monty` is a `napi` binding over the Rust `monty-pool`; exposes a `Monty` pool with `checkout().feedRun(...)`, streamed prints, async external functions, and `MontyCrashedError`. Platform npm packages ship both the `.node` library and the `monty` binary; the client auto‑resolves or accepts `binaryPath` (PATH executability check + 10s startup timeout). A wasm‑only in‑process API is available at `@pydantic/monty/wasm`. Hardening/bug fixes: non‑enumerable tuple markers, kwargs/dict objects use null‑prototype records and `defineProperties` to block `__proto__` pollution, paused readers restore backpressure, improved binary/platform package handling, half‑open i64 bound fix, tracebacks rendered once in Rust and shipped as a string, and updated smoke tests (compile TS and run JS).
  - CI: enforces proto sync (`make check-proto`), builds a real `monty` binary for protocol/pool tests, publishes platform npm packages (napi + binary + wasm), fixes Node 20/Windows CI, safer smoke‑test packaging (installs the CLI wheel and runs a pool smoke test), restores Windows static CRT linking for the `.node`, and cleans up Python tests.
  - Benchmarks: adds `monty-bench` subprocess‑pool benches (`pool_create_session_run`, `session_checkout_run`, `ext_calls_1000`) and wires them into CodSpeed; CI builds the `monty` worker in release and runs all benches with `MONTY_TEST_BIN` set so workers don’t build mid‑run.
  - Docs: tightened docstrings and trimmed comments across crates for clarity.
  - monty: invalid REPL inputs now raise `MontyRuntimeError` consistently (non‑str keys → `TypeError`, lone‑surrogate strings → `ValueError`).

- **Migration**
  - Python: replace direct `.run()`/`.run_async()` with session calls.
    - Before: `m = pydantic_monty.Monty(code, ...) ; m.run(...)`
    - Now: `with Monty() as pool: with pool.checkout(...) as s: s.feed_run(code, inputs=..., external_functions=...)`
    - Async: `async with AsyncMonty() as pool: async with pool.checkout() as s: await s.feed_run(...)`
  - Node.js: use the pool API; implementation is now native (`napi`) but the public API stays the same. For in‑process execution (e.g. browsers), use `@pydantic/monty/wasm`.
    - Before: `new Monty(code, opts).run({ inputs, externalFunctions })`
    - Now: `const pool = await Monty.create(); const s = await pool.checkout({...}); const out = await s.feedRun(code, { inputs, externalFunctions, printCallback });`
    - The `monty` binary is resolved via platform npm packages; override with `MONTY_BIN`/`binaryPath` if needed. Script metadata and type‑checking move to `checkout(...)` args. Legacy REPL/snapshot APIs and old serialization modules were removed in favor of pool sessions.

<sup>Written for commit 7bf1cda720b78543576a06e39efb6dbb3c50330c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/500?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


samuelcolvin:
@cubic-dev-ai review

samuelcolvin:
@cubic-dev-ai update the PR description

strawgate:
@coderabbitai do a thorough review

.github/workflows/ci.yml
```diff
@@ -66,6 +66,10 @@ jobs:
         env:
           SKIP: no-commit-to-branch
 
+      # the protobuf schema and the checked-in prost output must never drift
+      - name: Check generated protobuf code is in sync
+        run: make check-proto
```
davidhewitt, line 71: Should this be part of the pre-commit config?

.github/workflows/ci.yml
```diff
@@ -112,6 +116,13 @@ jobs:
       - run: cargo llvm-cov run --no-report -p monty-datatest --features ref-count-return
       # coverage for `make test-type-checking`
       - run: cargo llvm-cov --no-report -p monty_type_checking -p monty_typeshed
+      # subprocess protocol + child mode + worker pool. The pool tests spawn a
+      # real `monty` binary: build it plainly first (outside the llvm-cov
+      # target dir) and point the tests at it explicitly.
+      - run: cargo build -p monty-cli
+      - run: cargo llvm-cov --no-report -p monty-proto -p monty-cli -p monty-pool
```
davidhewitt, line 123: Seems questionable to build `monty-cli` outside of the tests in this way, we won't get coverage?

Might be able to use `llvm-cov show-env` to get configuration for instrumenting the cli for tests. We do this below for the python code.

.github/workflows/ci.yml
```diff
@@ -161,6 +172,11 @@ jobs:
         # don't use .venv python in CI
       - run: rm .cargo/config.toml
 
+      # pydantic-monty depends on pydantic-monty-cli; install it from the
+      # workspace so `maturin develop` can resolve it (plain build, outside
+      # the llvm-cov env, so the worker binary is not instrumented)
```
davidhewitt, line 177: why do we want it not instrumented?

.github/workflows/codspeed.yml
```diff
@@ -41,11 +41,25 @@ jobs:
...
+        run: cargo codspeed build -p monty-bench
 
       - name: Run benchmarks
         uses: CodSpeedHQ/action@d872884a306dd4853acf0f584f4b706cf0cc72a2 # v4.13.0
         with:
           mode: simulation
-          run: cargo codspeed run -p monty-bench --bench main
+          run: cargo codspeed run -p monty-bench
+        env:
+          MONTY_TEST_BIN: ${{ github.workspace }}/target/release/monty
```
davidhewitt, line 65: I think for the pool benchmarks it would make more sense if there was a single binary that internally spawned a protobuf service and client as threads and called between the two. Otherwise the runtime side of the pool will be opaque to codspeed and it'll pretty much kill the simulation benefits.

crates/monty-bench/benches/pool.rs
davidhewitt, line 1: see comment in `codspeed.yml`.

crates/monty-cli/src/subprocess.rs
```diff
@@ -0,0 +1,886 @@
...
+/// `ReplProgress` (suspended). The session meta carries the child-side state
+/// that lives *outside* the repl — script name and accumulated type-check
+/// stubs — so a `Load`ed session keeps type-check enforcement instead of
+/// silently dropping it:
+///
+/// `[script_name str][type_check u8]` and, when `type_check` is 1,
+/// `[committed_stubs str][has_pending u8][pending_snippet str?]`, where each
+/// `str` is a `u32 LE` byte length followed by UTF-8 bytes. The payload is
+/// monty's postcard format — only a monty child of the same version can
+/// restore it.
```
davidhewitt, line 46: Is this comment especially about `postcard` out of date given it's now protobuf?

crates/monty-cli/src/subprocess.rs
```diff
@@ -0,0 +1,886 @@
...
+    }
+}
+
+/// Writes one event frame to stdout.
+///
+/// Framing is stateless and `Stdout` handles share one global buffer, so a
+/// fresh handle per write is safe.
+fn send(event: &pb::ChildEvent) -> Result<(), monty_proto::FrameError> {
+    write_frame(&mut io::stdout(), event)
+}
```
davidhewitt, line 601: I think eventually we might want to support http endpoint (websocket) for the protobuf connection rather than stdout/stdin, probably a follow up?

crates/monty-cli/src/subprocess.rs
```diff
@@ -0,0 +1,886 @@
...
+                        match &tc.pending_snippet {
+                            Some(snippet) => {
+                                state.push(1);
+                                push_str_field(&mut state, snippet);
+                            }
+                            None => state.push(0),
+                        }
+                    }
+                    None => state.push(0),
+                }
```
davidhewitt, line 373: This seems very imperative; I think we probably want a struct to declare this (serde with postcard, if it's opaque to the proto?)

crates/monty-proto/src/wire.rs
```diff
@@ -0,0 +1,1256 @@
+//! Hand-written [`prost::Message`] implementation for [`MontyObject`].
+//!
+//! Values are the hot payload of the protocol — every external function call
+//! ships its arguments, result, and final value across the process boundary.
+//! Generated prost code would force a mirror struct tree (`pb::MontyObject`)
+//! plus a deep conversion in each direction: a full clone of every string and
+//! container on encode, and a second tree rebuild on decode. Instead, the
+//! codegen maps the `monty.v1.MontyObject` schema message to [`WireObject`]
+//! via `extern_path` (see `src/bin/generate.rs`), and this module implements
+//! the wire format directly on top of [`MontyObject`]:
```
davidhewitt, line 10: Do we have benchmarks to prove that this is necessary instead of just implementing `From` / `Into` conversions to generated prost messages? I'm also not entirely convinced by the cloning argument, might be possible to move data without cloning in those trait implementations.

Seems like a lot of code.

crates/monty-proto/src/frame.rs
```diff
@@ -0,0 +1,196 @@
...
+
+use prost::Message;
+
+use crate::wire::reset_decode_budget;
+
+/// Default maximum frame length (256 MiB).
+///
+/// Far above any sane payload, but small enough that a corrupted length
+/// prefix cannot trigger a multi-gigabyte allocation in the receiver.
+pub const MAX_FRAME_LEN: u32 = 256 * 1024 * 1024;
```
davidhewitt, line 26: Should this be configurable?

crates/monty-python/example.py
davidhewitt, line 1: It seems like we are breaking the existing Python API pretty hard.

Could we potentially make it such that if a user uses the existing APIs then we create a pool internally for them? (and maybe deprecate those for a few versions.)

Thinking for existing users of monty, possibly even pydantic AI?

crates/monty-js/README.md
davidhewitt, line 1: Would you be open to splitting off the JS changes into a follow-up PR? We could then get the changes right in the Python & Rust APIs first, keeping the JS one in-process in the short term. Smaller review diff and less churn here while iterating.

flyingmutant:
From e5f9e89:

> A monty process can never be made fully crash-proof against memory errors
(stack overflow aborts, allocator aborts) triggered by adversarial input.

Can you please expand on this? I though the whole premise of Monty was to achieve safety for adversarial input without extra isolation boundaries like subprocesses.

If we are using subprocess isolation to protect against adversarial input, OS-level sandboxing for the process is readily available and very robust – and at that point, regular Python can be used.

README still says

> Startup extremely fast (<1μs to go from code to execution result)

Is direct Monty usage (without `monty-pool`) considered safe and appropriate for production?

samuelcolvin:
> Startup extremely fast (<1μs to go from code to execution result)

Should be updated to <50μs.

> Can you please expand on this? I though the whole premise of Monty was to achieve safety for adversarial input without extra isolation boundaries like subprocesses.

Preventing all possible panics and memory errors, including in libraries we use (e.g. ruff, ty) is basically impossible - see https://github.com/astral-sh/ruff/pull/25464. Using subprocesses makes these failures much less severe - the worker running that code dies and can be replaced in around 20ms.

Also using subprocesses means:
* we can use a custom allocator to guarantee memory limits
* we can kill the process if cooperative time checks are somehow skipped
* a single process or server can run a pool of workers and run many Monty instances in parallel

> If we are using subprocess isolation to protect against adversarial input, OS-level sandboxing for the process is readily available and very robust – and at that point, regular Python can be used.

In a way that's OS, OS version, architecture and setup agnostic and can reliably work when someone is installing a package from PyPI or NPM? Dream on.

PR #520, opened 2026-06-30
rename and change behaviour of `external_lookup`

samuelcolvin:
## Summary by cubic
Renamed the external function hook to external lookup and expanded it: undefined names now resolve lazily to host functions or plain values across Python, JS pool sessions, and the `@pydantic/monty/wasm` in‑process path.

- New Features
  - Added `external_lookup`/`externalLookup` for lazy name resolution: callables become host functions; other values are returned on read; absent names raise `NameError`. `inputs` wins when a name exists in both; function proxies are cached; value reads are not cached in CPython.
  - Safety/consistency: JS and WASM resolve/invoke own keys only; falsy values and `null`/`undefined` resolve; calling a cached proxy after its entry becomes non‑callable now raises a CPython‑style `TypeError`; dict access errors propagate; Python type objects round‑trip as types. WASM restores the mount table if a lookup/external call errors mid‑run.
  - API flow: `feed_start`/`feedStart` accept `external_lookup`; snapshots add `resume_auto`/`resumeAuto` to auto‑drive lookups, function calls, and futures. Conversion errors surface and end the session.

- Migration
  - Replace `external_functions` → `external_lookup` in Python (`pydantic_monty`) and `externalFunctions` → `externalLookup` in JS (`@pydantic/monty` and `@pydantic/monty/wasm`).
  - Use callables for calls; supply plain values for bare‑name reads. Missing names raise `NameError`; bad value conversions raise and end the session.

<sup>Written for commit e6bdcd070475d20912391c499dca2eba2542ae8c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/520?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -712,7 +712,7 @@ await using session = await pool.checkout({ typeCheck: false })
 await session.feedRun('x = 21') // session state persists across feeds
 const result = await session.feedRun('x * 2', {
   inputs: { y: 1 },
-  externalFunctions: { fetch: async (url: string) => '...' }, // sync or async
+  externalLookup: { fetch: async (url: string) => '...' }, // sync or async
   printCallback: (stream, text) => {},
 })
 ```
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ async def main():
             output = await session.feed_run(
                 code,
                 inputs={'prompt': 'testing'},
-                external_functions={'call_llm': call_llm},
+                external_lookup={'call_llm': call_llm},
             )
     print(output)
     #> example output, message count 2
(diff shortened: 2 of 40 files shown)


davidhewitt:
Is this intended to (eventually) replace `dataclass_registry`? I imagined we might want to have "proxy" objects where the Monty VM delegates all attribute lookup, function calls etc to the host.

Seems like accepting values here has a lot of overlap with dataclasses.

samuelcolvin:
> Is this intended to (eventually) replace `dataclass_registry`? I imagined we might want to have "proxy" objects where the Monty VM delegates all attribute lookup, function calls etc to the host.
> 
> Seems like accepting values here has a lot of overlap with dataclasses.

I want to remove the dataclass registry somehow, but not sure this is the right way, I guess it might be.

PR #525, opened 2026-07-03
use web workers for browser wasm

davidhewitt:
This PR changes the wasm code to also use "out-of-process" execution by using web workers for the monty execution.

Fully vibe coded by Claude, I need to give this a careful review to make sure that it matches the API we actually want.

---
## Summary by cubic
Moves browser/wasm execution to a Web Worker pool backed by a lean `wasm32-wasip1` module, unified with native via the shared `monty-worker` state machine. Browsers now use a worker-backed backend with a hard kill; Node stays on the native pool; the CLI/runtime drive `monty_worker::Child`.

- New Features
  - `@pydantic/monty/wasm` exposes the worker API: `Monty.create()` auto-loads the bundled wasm; `createWorkerPool(module)` accepts a compiled module; falls back to in-process wasm if no worker is available.
  - Root `@pydantic/monty` resolves to the worker backend in browsers and the native pool on Node; new `@pydantic/monty/node` export exposes Node-only helpers (e.g. `MountDir`, snapshots).
  - New Rust crates: `monty-worker` (transport-agnostic child state machine) and `monty-wasm-runtime` (WASI reactor exporting `monty_dispatch_turn`); `monty-runtime` is a thin shell over `monty-worker`.
  - TypeScript worker transport/pool with a minimal protobuf codec and value mapping; browser uses `@bjorn3/browser_wasi_shim` as the WASI host; Node threads supported via `nodeWorkerFactory(module)`.
  - `MontyRuntimeError.display('traceback')` now renders tracebacks from frames when no pre-rendered text is present.
  - Unified Vitest suite for Node and browser, adds public API conformance tests, and stubs Node builtins in browser tests.

- Migration
  - Build the wasm module with `cargo build -p monty-wasm-runtime --target wasm32-wasip1 --release`; `npm run build:wasm` copies the asset next to the loaders.
  - `@pydantic/monty/wasm` now exports the worker pool; use `Monty.create()` or `createWorkerPool(module)`. The legacy in-process `wasm` surface and `wasm_*` tests were removed.
  - Import Node-only helpers (mounts, snapshots) from `@pydantic/monty/node`; the wasm worker rejects filesystem mounts in browsers.
  - Package exports use `browser`/`node` conditions; the `@pydantic/monty-wasm32-wasi` platform package and legacy scripts/targets were removed.

<sup>Written for commit 1906c7b9412dfe76552dc94fc261d6c3419bdac8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/525?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty-runtime/Cargo.toml
```diff
@@ -20,7 +20,7 @@ clap = { workspace = true }
 monty = { workspace = true }
 monty-proto = { workspace = true }
 monty-type-checking = { workspace = true }
-prost = { workspace = true }
+monty-worker = { path = "../monty-worker" }
```
samuelcolvin: if we're publishing packages, we can't do this.

crates/monty-cli/src/subprocess.rs
```diff
@@ -1,73 +1,47 @@
-//! `monty subprocess`: protocol child mode.
+//! `monty --subprocess`: protocol child mode.
```
samuelcolvin: this is wrong (or an unnecessary change)

samuelcolvin (commented):
Overall looks great, and I'm a big fan of the deduplication.

I'm worried it's behind main as it was talking about `--subprocess`.

davidhewitt:
Will address the reviews in a followup so this doesn't block the b3 release.

PR #528, opened 2026-07-04
small perf wins

samuelcolvin:
Numerous small optimizations.

---
## Summary by cubic
Optimized hot interpreter paths and string conversion by lazily resolving traceback locations, pooling frame locals, interning singleton `repr`s, speeding up `int` formatting and `str(int)`, adding a no-kwarg call fast path, and returning heap `str` `Value`s from `py_str`/`py_repr` with a faster `Value::to_str`. Bench: fib ~-14%, list_append_str ~-38%; no user-visible changes.

- **Performance**
  - Lazy traceback lookup: store `call_offset: Option<u32>` and resolve during unwind (`resolve_offset`); snapshots/async carry offsets.
  - Pooled frame locals: build in `namespace_scratch` under `HeapGuard`, then `append` to the stack; reuses the buffer across calls.
  - Faster ints: `Value::Int` `repr` via `itoa::Buffer`; `str(int)` formats via `itoa` in `Str::init` to a single right-sized allocation.
  - No-kwarg call fast path: handle simple signatures before constructing/guarding kwargs iterators.
  - String conversion: `py_str`/`py_repr` now return heap `str` `Value`s; added `Value::to_str`/`to_str_heap` to resolve without clones and avoid borrow conflicts; used in `print`, f-strings, `format`, and error paths.
  - Interned singleton `repr`s: `None`/`True`/`False`/`Ellipsis` return interned `str` values with zero allocation.
  - Lazy repr cycle guard: `LazyHeapSet` allocates only when needed.

- **Dependencies**
  - Added `itoa` to the workspace and `crates/monty`.

<sup>Written for commit 2148005297d26e389d03ec81c9d86c3bab75bc1b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/528?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/exception_private.rs
+++ b/crates/monty/src/exception_private.rs
@@ -397,7 +397,11 @@ impl ExcType {
             Ok(key_value) => {
                 // `key_value` is a heap `str` `Value`; extract its text and drop it.
                 defer_drop!(key_value, vm);
-                key_value.to_str(vm).unwrap_or_default().to_owned()
+                if let Ok(s) = key_value.to_str(vm) {
+                    s.to_owned()
+                } else {
+                    format!("<{}>", key.py_type_name(vm))
+                }
             }
             Err(_) => format!("<{}>", key.py_type_name(vm)),
         };
--- a/crates/monty/src/exception_private.rs
+++ b/crates/monty/src/exception_private.rs
@@ -394,7 +394,11 @@ impl ExcType {
     /// `KeyError` is always raised rather than a spurious `ValueError`.
     pub(crate) fn key_error(key: &Value, vm: &mut VM<'_, impl ResourceTracker>) -> RunError {
         let key_str = match key.py_str(vm) {
-            Ok(s) => s.into_owned(),
+            Ok(key_value) => {
+                // `key_value` is a heap `str` `Value`; extract its text and drop it.
+                defer_drop!(key_value, vm);
+                key_value.to_str(vm).unwrap_or_default().to_owned()
+            }
             Err(_) => format!("<{}>", key.py_type_name(vm)),
         };
         SimpleException::new_msg(Self::KeyError, key_str).into()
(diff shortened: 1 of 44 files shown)


crates/monty/src/types/str.rs
```diff
@@ -69,6 +69,10 @@ impl Str {
         let StrInitArgs { object } = StrInitArgs::from_args(args, vm)?;
         match object {
             None => Ok(Value::InternString(StaticStrings::EmptyString.into())),
+            // Fast path: `str(int)` formats via itoa straight into a
+            // right-sized allocation, skipping the grow-then-shrink `String`
+            // that the generic `py_str`/`py_repr` path builds and then reboxes.
+            Some(Value::Int(i)) => Ok(allocate_string(itoa::Buffer::new().format(i), vm.heap)?),
```
davidhewitt: I don't think we need this here as it's already added in `py_repr_fmt`.

davidhewitt (approved):
Generally lgtm

crates/monty/src/types/str.rs
samuelcolvin: It gives a 38% speed improvement on `list_append_str` - not sure if this is the string stuff in the comment, the drop stuff or creating the `HashSet` that is created and never used.

samuelcolvin: fixed.

crates/monty/src/exception_private.rs
```diff
@@ -394,7 +394,11 @@ impl ExcType {
     /// `KeyError` is always raised rather than a spurious `ValueError`.
     pub(crate) fn key_error(key: &Value, vm: &mut VM<'_, impl ResourceTracker>) -> RunError {
         let key_str = match key.py_str(vm) {
-            Ok(s) => s.into_owned(),
+            Ok(key_value) => {
```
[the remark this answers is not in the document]
samuelcolvin: fixed

PR #529, opened 2026-07-05
Bound native `evaluate_function` re-entry to prevent infinite recursion

samuelcolvin:
A recursive Python callback evaluated through `VM::evaluate_function` (map/filter/ sorted key callbacks, __repr__/__str__ dispatch, class-valued __init__ cycles) re-entered the Rust interpreter loop on the native call stack with no depth limit, overflowing the stack and aborting the whole process before Monty's Python-level recursion limit could ever trip. This is fatal for the in-process wasm API, which shares the host process.

Adds a second, much lower recursion budget (MAX_RUN_REENTRY_DEPTH) dedicated to native re-entry, mirroring the existing RecursionGuard machinery, charged at evaluate_function's entry so it covers both the framed (map/filter/sorted/ __repr__) and frameless (class-valued __init__) re-entry cycles. The constant was tuned empirically by bisecting real SIGABRT depths rather than guessed.

Claude-Session: https://claude.ai/code/session_01MLmyMBZFoGpBzQzRjjMxah

---
## Summary by cubic
Bound native re-entry in `VM::evaluate_function` to prevent stack overflows. Recursive callbacks now raise `RecursionError` instead of crashing, including frameless class‑valued `__init__` cycles.

- **Bug Fixes**
  - Added fixed cap `MAX_RUN_REENTRY_DEPTH`, enforced via `run_reentry_depth` and `RunReentryGuard`, charged at `evaluate_function` entry to cover both framed callbacks and frameless `__init__` cycles.
  - Exceeding the cap raises `RecursionError` for recursive `map`/`filter`/`sorted(key=...)` and `__repr__`/`__str__`; modest finite chains still work.
  - `run_reentry_depth` resets on restore and is never serialized; `snapshot` asserts we’re not inside re-entry.
  - Added tests for nested eval, recursive `__repr__`/`__str__`, a finite chain, and the `A.__init__ = A` regression; updated `limitations/*` to document earlier recursion errors and test-only, lowering‑only `sys.setrecursionlimit`.

<sup>Written for commit 10885641fbefac5e5ba4d6fd35e95b08be9460e7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/529?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/bytecode/vm/call.rs
+++ b/crates/monty/src/bytecode/vm/call.rs
@@ -372,7 +372,7 @@ impl<T: ResourceTracker> VM<'_, T> {
     /// execute them and resume, which this synchronous context cannot support.
     ///
     /// The nested `self.run()` below recurses on the native Rust stack, so
-    /// re-entry is bounded via [`incr_run_reentry`](Self::incr_run_reentry)
+    /// re-entry is bounded via [`enter_run_reentry`](Self::enter_run_reentry)
     /// at entry — before `call_function`, since a class-valued `__init__` can
     /// recurse back in without ever pushing a frame.
     pub(crate) fn evaluate_function(
@@ -381,7 +381,7 @@ impl<T: ResourceTracker> VM<'_, T> {
         callable: &Value,
         args: ArgValues,
     ) -> Result<Value, RunError> {
-        if let Err(e) = self.incr_run_reentry() {
+        if let Err(e) = self.enter_run_reentry() {
             // Bailing before `call_function` takes ownership of `args`, so
             // reclaim its refcounts here.
             args.drop_with_heap(self);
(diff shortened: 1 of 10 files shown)


davidhewitt:
See also #525 which will help with the wasm case (and make a lot of the comments redundant)

samuelcolvin:
#525 looks great, but that's different to this - I agree I need to fix the comments.

davidhewitt:
yep I think both make sense

PR #538, opened 2026-07-07
Add `decimal` module (#218)

### LINKED ISSUE #218: decimal module support in scope?
Is the standard decimal module (for Decimal type) currently supported or planned?
We're using Monty for financial formulas where float precision is a dealbreaker. Right now, we have to wrap all math logic in host-side functions to enforce Decimal usage.

imran2140:
Closes #218

## Summary
Implements `decimal.Decimal` as a Rust port of CPython 3.14's `_pydecimal.py` numeric core, running under a **fixed context** matching CPython's defaults (`prec=28`, `ROUND_HALF_EVEN`, `Emax=±999999`). Arithmetic results are digit-for-digit identical to CPython's — the ported functions cite their `_pydecimal`  line numbers so they can be diffed against the original.

Divergences from CPython (fixed context, reduced method surface, input caps, qualified `__name__`s) are documented in `limitations/decimal.md`.

Assisted-by: Claude:claude-5-fable

## Implementation notes
- **Numeric core** (`crates/monty/src/types/decimal/`): construction from  `str | int | float | Decimal` and the `as_tuple` form; `add`/`sub`/`mul`/`div`, `floordiv`/`mod`/`divmod`, 2- and 3-arg `pow`, `sqrt`/`ln`/`log10`/`exp`,  `quantize`/`normalize` `to_integral_value`, the `is_*` predicates, `as_tuple`/`adjusted`/`copy_*`, and the builtins `int()`/`float()`/`round()`/`abs()`/`divmod()`/`pow()`/`hash()` plus `math.floor`/`ceil`/`trunc`.
- **Sandbox guards**: CPython's algorithms are bounded *given* bounded operands, so every constructor path caps coefficients/NaN payloads at 4300 digits (the `int`↔`str` limit) and bounds literal exponents; attacker-scalable allocations (powers of ten, format zero-pads) are charged to the `ResourceTracker`, raising `MemoryError` where CPython would build a multi-GB value.
- **Dispatch**: a single `decimal::binary_op_value` probe serves all seven binary operators and the `pow()` builtin — promotion runs before the coefficient is cloned, and the probe owns CPython's sequence-repeat special case (`'a' * Decimal(2)`).
- **Comparisons/hash**: exact CPython semantics including NaN/sNaN (the VM's identity shortcut now exempts Decimal so `nan != nan` even for the same object); `hash(Decimal(5)) == hash(5)` cross-type consistency in-sandbox.
- **Exceptions**: the full taxonomy is importable with CPython's multi-parent relationships;  only the default-trapped signals (`InvalidOperation`, `DivisionByZero`, `Overflow`) ever raise.
- **Formatting**: `str`/`repr` and the full `__format__` spec mini-language, emitted through tracker-reserved `StringBuilder`s.
- **VM/heap/snapshots**: `HeapData::Decimal` + `HeapReadOutput` variants (appended last for snapshot index stability, like `Type`/`MontyObject`/`MontyType`); snapshot deserialization validates Decimal payloads and rejects corrupt bytes safely.
- **Host boundary**: a `Decimal` crosses losslessly as its canonical string — `MontyObject::Decimal` through the wire protocol (proto tag appended), `decimal.Decimal` in `pydantic_monty`, a tagged string in `@pydantic/monty`; hand-authored typeshed stub advertises only the implemented surface.
- **Tests**: seven `test_cases/decimal__*.py` files assert exact values and exception messages under both CPython and Monty; `decimal_guards.rs` covers the resource-limit paths; binary/JSON serde, host-input, and Python/JS binding tests cover the boundary.

---
## Summary by cubic
Adds a `decimal` module implementing `decimal.Decimal` under a fixed context matching CPython defaults (prec=28, `ROUND_HALF_EVEN`, Emax=±999999). Results, formatting, comparisons, and exceptions match CPython; values cross the host boundary losslessly as canonical strings.

- New Features
  - `decimal.Decimal` port with constructors (str/int/float/Decimal/as_tuple), arithmetic (`+ - * / // % divmod pow` incl. 3-arg), transcendentals (`sqrt ln log10 exp`), methods (`quantize normalize to_integral_value copy_sign as_tuple`), predicates, and `__format__`; builtins `int/float/round/abs/divmod/pow/hash` and `math.floor/ceil/trunc` wired up.
  - Exact CPython comparison, NaN/sNaN behavior, and cross-type hash (`hash(Decimal(5)) == hash(5)`); VM equality operator updated to honor Decimal semantics.
  - Full `decimal` exception hierarchy (`decimal.DecimalException`, `InvalidOperation`, `DivisionByZero`, `Overflow`, etc.); only the default-trapped signals raise.
  - Host boundary integration:
    - New `MontyObject::Decimal` carried as the canonical string (e.g. `"1.20"`, `"1E+5"`, `"NaN"`); Protobuf adds field 29; JSON uses externally tagged form `{"Decimal":"<str>"}`.
    - JS: `{ __monty_type__: 'Decimal', value: '<str>' }` round-trips; Python: `Decimal` converts via `str()`; bindings updated in `pydantic_monty` and `@pydantic/monty`.
    - Snapshots/serde support Decimal and validate payloads on load.
  - Sandbox guards: coefficient/NaN payload capped at 4300 digits; literal exponent bounds enforced; large `10**k` and format zero-pads are charged to the resource tracker; oversized boundary inputs are rejected.
  - Typeshed: added `decimal.pyi` subset (fixed-context surface) and enabled in version tables; type-checking tests cover common operations.
  - Stdlib: `decimal` is now importable; README updated.

- Migration
  - If you consume Monty objects over RPC/JSON, add handling for `MontyObject::Decimal` (Protobuf tag 29; JSON `{"Decimal":"<str>"}`) and, in JS, `{ __monty_type__: 'Decimal', value: '<str>' }`.
  - Exception mapping: hosts should recognize `decimal.*` exception names so `except decimal.*` in-sandbox catches mapped host errors.
  - No snapshot incompatibility: new enum/variant entries are appended to preserve existing indices.

<sup>Written for commit 944dd528e277931bc243a0fb9259d41dcc33a024. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/538?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty-typeshed/vendor/typeshed/stdlib/decimal.pyi
davidhewitt, line 1: this file is almost certainly not vendored, looks suspicious?

crates/monty/src/value.rs
```diff
@@ -1598,6 +1730,29 @@ impl Value {
...
+            && id == other_id
+        {
+            match vm.heap.read(*id).py_eq_impl(other, vm)? {
+                Some(result) => Ok(result),
+                None => Ok(true), // default object identity
+            }
+        } else {
+            self.py_eq(other, vm)
+        }
+    }
```
davidhewitt, line 1754: Please revert this split; instead we should fix `py_eq_impl` to not have the short-circuit.

crates/monty/src/value.rs
davidhewitt, line 1: The changes in this file are pretty gross; we should probably clean up the trait so that these changes aren't necessary and they are more localised to the `decimal` implementation.

crates/monty/src/run.rs
```diff
@@ -123,7 +123,7 @@ impl MontyRun {
     /// # Errors
     /// Returns an error if deserialization fails.
     pub fn load(bytes: &[u8]) -> Result<Self, postcard::Error> {
-        postcard::from_bytes(bytes)
+        from_snapshot_bytes(bytes)
```
davidhewitt, line 126: This general set of changes seems suspect, should either be split out and justified separately or reverted.

crates/monty/src/types/decimal/mod.rs
davidhewitt, line 1: This seems problematic; why not use a rust implementation such as `rust-decimal`?

davidhewitt (changes requested):

crates/monty-typeshed/vendor/typeshed/stdlib/decimal.pyi
imran2140: Yep, it's duplicate of `crates/monty-typeshed/custom/decimal.pyi`, the duplication seems intended according to the crate's [`README`](https://github.com/pydantic/monty/tree/46d12174acd30e108f263bc7ffe25e75bb6e8b2d/crates/monty-typeshed/custom). (copied by [`monty-typeshed/update.py#L332-L337`](https://github.com/pydantic/monty/blob/46d12/crates/monty-typeshed/update.py#L332-L337), introduced in https://github.com/pydantic/monty/pull/48)

crates/monty/src/types/decimal/mod.rs
imran2140: I initially tried to base the implementation on [`fastnum`](https://github.com/neogenie/fastnum/), which seemed like a good fit given it tries to follow [IBM’s General Decimal Arithmetic Specification](https://speleotrove.com/decimal/decarith.html), like python stdlib [`decimal`](https://github.com/python/cpython/blob/v3.14.6/Lib/decimal.py#L6).  But Claude kept layering on workarounds for  various operations because of subtle differences in the observed behavior in tests (e.g. `Decimal('-0') // Decimal('1')` produces `Decimal('-0')` in Python preserving the sign, which wasn't the case with the out of the box implementation in `fastnum`).  I also had implementation of decimal Context in-scope in that fastnum based port, which compounded the complexity.  When I decided to re-implement again, I thought letting Claude port limited part of `_pydecimal` may not be a bad idea, and it also avoided the need to bring in another dependency.

Let me know if you'd prefer me to open an alternative PR with the `fastnum` based implementation, or iterate on this PR to make it merge-able.

PR #543, opened 2026-07-09
Add cargo-nextest for Rust tests

zozo123:
## Summary
- Add a minimal nextest config with `fail-fast = false` so Rust test runs report the full failure set.
- Switch regular Rust `cargo test`-style Makefile commands and the macOS/Windows Rust CI matrix to `cargo nextest run`.
- Keep Linux coverage on `cargo llvm-cov`'s default test runner after the first PR CI run showed `cargo llvm-cov nextest` was slower there; custom datatest, doctest, miri, Python, JS, and wasm runners stay unchanged.

## Test plan
- `make -n test-no-features test-type-checking test-subprocess testcov`
- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci.yml"); puts "workflow yaml ok"'`
- Not run locally: `cargo nextest ...` and `cargo llvm-cov ...` because those cargo subcommands are not installed in this environment.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,2 @@
+[profile.default]
+fail-fast = false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,10 +106,6 @@ jobs:
         with:
           tool: cargo-llvm-cov
 
-      - uses: taiki-e/install-action@1ed3272338f573e042a2e6bca3893aa19f43b47a # v2.71.3
-        with:
-          tool: nextest
-
       # this means pyo3 will use Python 3.14 in tests
       - name: set up python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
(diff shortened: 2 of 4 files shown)


zozo123:
Timing note from the first CI run: routing the Linux coverage job through `cargo llvm-cov nextest` was mixed-to-negative (`test-rust` was ~11:02 on the PR vs ~7:56 on the latest `main` run). I pushed c3dca54 to keep coverage on the existing `cargo llvm-cov` path and limit nextest to regular Rust test runs, where the isolation/reporting benefits are clearer and the coverage slowdown risk is avoided.

zozo123:
Follow-up after the conservative revision (`c3dca54`): all checks are green and the PR is merge-clean. The refined run keeps Linux coverage on plain `cargo llvm-cov` and finished `test-rust` in ~7:24, back in line with recent `main` runs (~7:48-7:56). The nextest OS-matrix jobs also passed: macOS ~3:18 and Windows ~4:15. So the final shape avoids the coverage slowdown from the first attempt while preserving nextest where it gives cleaner Rust test isolation/reporting.

davidhewitt:
Thanks for the PR. I don't really see why we need the complexity, especially if `cargo test` can't be fully replaced?

zozo123:
That's fair — I dug into the numbers further and agree this doesn't earn its complexity. The CI data shows no real wall-clock win (the nextest steps ran at or slightly above the current `cargo test` times on macOS/Windows, and the coverage path was slower), doctests would still need `cargo test`, and datatest/miri/coverage stay on their own runners regardless — so it ends up as a second partial test path rather than a replacement. Closing — thanks for taking a look!

PR #549, opened 2026-07-10
Monty-proto moves

samuelcolvin:
Two changes:

* moves python conversion logic into `monty-proto` - to avoid `monty-cpython` depending on `monty-python`, and `monty-python` needing `crate-type = ["cdylib", "rlib"]`
* removes `monty-worker` and moves the logic into `monty_proto::worker`

---
## Summary by cubic
Moved the Python ↔ MontyObject conversion layer and the worker state machine into `monty-proto` to simplify dependencies, cut linking overhead, and remove `monty-worker`. `monty-python` stays a pure `cdylib`; consumers enable `monty-proto` features instead.

- **Refactors**
  - Moved conversion code (`convert`, `dataclass`, exception mapping) into `monty-proto::python` behind a new optional `python` feature.
  - Updated `monty-cpython` to depend on `monty-proto` with `python` and drop the `_monty` rlib dependency.
  - `monty-python` reuses the shared layer via `monty-proto::python`, remains `crate-type = ["cdylib"]`, and re-exports `MontyFileHandle`.
  - Moved the protocol child state machine into `monty_proto::worker` behind a `worker` feature; removed `monty-worker`. `monty-runtime` and `monty-wasm-runtime` now use `monty-proto` with `worker`.
  - Switched Py version checks in the shared code to runtime `py.version_info()`.
  - Centralized `pyo3`/`pyo3-build-config` pins in the workspace; the `python` feature links `pyo3` but never enables `extension-module`, so pure-Rust consumers stay pyo3-free.

- **Migration**
  - Embedders: add `monty-proto = { features = ["python"] }` and import from `monty_proto::python::{monty_to_py, py_to_monty_value, DcRegistry, exc_*}`.
  - Workers: replace `monty-worker` with `monty-proto = { features = ["worker"] }` and import from `monty_proto::worker::*`.

<sup>Written for commit cd4cc9a65f6a9aa9eb9b788172d3eee38ba814c0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/549?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty-proto/src/python/dataclass.rs
```diff
@@ -254,8 +254,10 @@ impl PyUnknownDataclass {
                 py.None().into_bound(py).get_type().into_any()
             };
 
-            // Create a Field object with the required attributes
-            let field_obj = if cfg!(Py_3_14) {
+            // Create a Field object with the required attributes.
+            // Runtime version check (not `cfg!(Py_3_14)`): this crate has no
+            // pyo3-build-config build script, so the version cfgs don't exist.
+            let field_obj = if py.version_info() >= (3, 14) {
```
samuelcolvin, line 260: @davidhewitt this is the only thing I think is worth checking - should we add config like to `build.rs`, or keep this change?

davidhewitt: This is probably fine, we're making a runtime decision here on some dynamic Python code, we don't need to conditionally compile.

But it does hint that if we add more complex python logic in this layer, we might want the build.rs. We'll have to unconditionally build `pyo3-build-config` but we could gate consuming it on enabling a pyo3 dependency.

PR #553, opened 2026-07-10
refactor PGO implementation

davidhewitt:
This changes the python wheel builds so that the `monty-python` client crate is no longer PGO-optimized, only the `monty-runtime` main binary.

At the same time I switch to use maturin's new `--pgo` flag to do the build.

There is a bug in maturin (not new in this PR) which I noted that the runtime is being built per-version; that shouldn't be necessary, I'll try to debug and fix upstream.

---
## Summary by cubic
Refactors PGO to optimize only the `monty-runtime` binary via `maturin --pgo` and folds runtime wheel builds into the main CI job. PGO now runs only on Linux x86_64, Linux aarch64, and macOS aarch64; Windows and other platforms build a runtime wheel without PGO. Training uses Python 3.14, and each platform produces a single Python-independent runtime wheel instead of one per Python version.

- **Refactors**
  - PGO applies only to `monty-runtime`; the `monty-python` extension builds normally.
  - `maturin --pgo` trains via `monty-datatest`'s new `pgo-exercise`, which runs `monty-pool` sessions against the interpreter test corpus.
  - The training pool shuts down cleanly so instrumented workers flush their profile data.
  - The interpreter flag is only passed on PGO platforms.
  - The Linux build installs Python 3.14 via `uv` inside the manylinux container for the training interpreter.
  - Removed `.github/actions/build-pgo-wheel`; artifacts are now `pypi_files-python-*` and `pypi_files-runtime-*`.
  - Updated `make dev-py-pgo` to install `llvm-tools` and build an installed PGO runtime wheel before developing `monty-python`.

- **Dependencies**
  - Bumped `maturin` to `>=1.15.0`, which fixes the per-version runtime wheel duplication.
  - Added `monty-pool` and `tokio` to `monty-datatest`; set `default-run` to `monty-datatest`.

<sup>Written for commit 2eec2623081add34dc7385b1a7c3d5a2d5cd2184. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/553?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


davidhewitt:
maturin issue reported as https://github.com/PyO3/maturin/issues/3256

Issue #483, opened 2026-06-02
MontyRepl time limit (max_duration_secs) is measured from REPL construction, not per feed_run


chkp-itamarst:
## Context

We're at Check Point, evaluating Monty as the sandboxed-Python backend for agents running against our systems — a long-lived `MontyRepl` per conversation with many `feed_run` snippets. We hit this in a real run.

## The issue

`max_duration_secs` isn't per-`feed_run`. It's a single wall-clock budget measured from when the `MontyRepl` is created and shared across every snippet (it even counts host time between calls), so once it elapses the REPL is permanently dead — every later snippet, even a trivial one, fails:

```python
repl = MontyRepl(limits=ResourceLimits(max_duration_secs=1.0))
repl.feed_run("t=0\nfor i in range(20_000_000): t+=i")  # ~0.8s -> OK
repl.feed_run("t=0\nfor i in range(20_000_000): t+=i")  # -> TimeoutError: time limit exceeded: 1.25s > 1.0s
repl.feed_run("print('hi')")                            # -> TimeoutError: time limit exceeded: ... > 1.0s
```

The cause is in `crates/monty/src/resource.rs`: `LimitedTracker.start_time` is set once at tracker creation and only reset by `set_max_duration()` / deserialize — `feed_run` takes no per-call limit and never resets it.

## Ask

Should `max_duration_secs` reset per `feed_run` for `MontyRepl` (or expose a per-`feed_run` limit)? If the cumulative behavior is intended?

davidhewitt:
It should be cumulative; this is definitely a bug.

FireMasterK:
Any updates on this? I would like to contribute a PR if this is open for contributions.

I would suggest either changing the behavior of `max_duration_secs` to be per feed, or maybe add a new field like `max_run_duration_secs`, or alternatively just change `max_duration_secs` to be a Union of float | `Timeout`, where timeout is a class which has a `total` and `per_run` field. What do the maintainers prefer?

davidhewitt:
I have been meaning to resolve this. We've adjusted monty's execution model to use a subprocess in the upcoming 0.19 release, which has affected some of the code in this area, when the dust settles on that change this is probably my top priority thing to clean up.

fugu-ai:
[...]

PR #554, opened 2026-07-10
use web-time for the resource-tracker clock so max_duration works on wasm32-unknown-unknown

Butch78:
`LimitedTracker` enforces `ResourceLimits::max_duration` via `std::time::Instant` (`crates/monty/src/resource.rs`). On `wasm32-unknown-unknown`, `Instant::now()` panics (`time not implemented on this platform`), so any run configured with a duration limit aborts. monty's wasm CI targets `wasm32-wasip1`, where `Instant` is implemented, so this path is green upstream and the gap is invisible there.

## Use case

We embed the `monty` crate directly as a `wasm32-unknown-unknown` wasm-bindgen module for in-process, browser-side Python in a cell (not the WASI/worker path). There, `max_duration` is the *only* limit that stops a non-allocating runaway such as `while True: pass` — `max_allocations` / `max_memory` never fire on it — and we can't arm it without panicking.

## Change

Swap the tracker clock to [`web_time::Instant`](https://crates.io/crates/web-time), a drop-in that re-exports `std::time::Instant` verbatim on native and WASI and is backed by `performance.now()` on `wasm32-unknown-unknown`. No behavior change on any currently-supported target.

## Verification

- Native: all resource-limit tests pass unchanged (incl. every `timeout_*` case and `suspension_time_does_not_count_toward_max_duration`).
- `wasm32-wasip1` (supported wasm target) and native compile clean; `clippy -D warnings` clean.
- `web_time::Instant` confirmed compiling for `wasm32-unknown-unknown` with monty's exact usage (`Instant::now()`, `Cell<Option<Instant>>`, `.elapsed()`).

I see the time code is in flux for the 0.19 subprocess model (#483) — happy to retarget or adjust timing to suit.

---
## Summary by cubic
Switch the resource tracker to `web-time::Instant` so `max_duration` works on `wasm32-unknown-unknown` without panicking. No behavior change on native or `wasm32-wasip1`; restores duration limits for browser-embedded `monty`.

- **Bug Fixes**
  - Replace `std::time::Instant` with `web_time::Instant` in the tracker to avoid `Instant::now()` panic on `wasm32-unknown-unknown`.
  - Enables reliable timeouts for non-allocating runaways (e.g., `while True: pass`) in browser builds.

- **Dependencies**
  - Add `web-time` v1.1 as a target-gated dependency for `wasm32-unknown-unknown`; import `web_time::Instant` only there and use `std::time::Instant` everywhere else.

<sup>Written for commit ba9ba9aedf3352566b10c81f2badffdbe26a3008. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/554?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2022,6 +2022,7 @@ dependencies = [
  "unicode-general-category",
  "unicode-normalization",
  "unicode_names2",
+ "web-time",
 ]
 
 [[package]]
--- a/crates/monty/Cargo.toml
+++ b/crates/monty/Cargo.toml
@@ -42,6 +42,12 @@ speedate = "0.17.0"
 itertools = "0.14.0"
 jiter = { version = "0.15.0", features = ["num-bigint"] }
 monty-macros = { workspace = true }
+# Portable monotonic clock: a drop-in for `std::time::Instant` that re-exports
+# std unchanged on native and WASI, and is backed by `performance.now()` on
+# `wasm32-unknown-unknown` (where `std::time::Instant::now()` panics). Keeps the
+# `max_duration` resource limit usable when `monty` is embedded directly in a
+# browser wasm module rather than run over WASI.
+web-time = "1.1"
 
 [features]
 # ref-count-return changes behavior to return information on reference counts to check they're correct
--- a/crates/monty/Cargo.toml
+++ b/crates/monty/Cargo.toml
@@ -42,11 +42,12 @@ speedate = "0.17.0"
 itertools = "0.14.0"
 jiter = { version = "0.15.0", features = ["num-bigint"] }
 monty-macros = { workspace = true }
-# Portable monotonic clock: a drop-in for `std::time::Instant` that re-exports
-# std unchanged on native and WASI, and is backed by `performance.now()` on
-# `wasm32-unknown-unknown` (where `std::time::Instant::now()` panics). Keeps the
-# `max_duration` resource limit usable when `monty` is embedded directly in a
-# browser wasm module rather than run over WASI.
+
+# `std::time::Instant` panics on `wasm32-unknown-unknown`, so the resource
+# tracker uses `web_time::Instant` (a `performance.now()`-backed drop-in) there
+# to keep `max_duration` working when `monty` is embedded directly in a browser
+# wasm module. Only that target needs it; every other target uses std directly.
+[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 web-time = "1.1"
 
 [features]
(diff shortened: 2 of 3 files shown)


davidhewitt:
This seems reasonable to me. Should we make it target-specific dependency just to avoid it where it's unnecessary?

I'm also curious about your use-case. It sounds like the work I just did in #525 will probably be relevant for you, which uses `wasip1` in a web worker to do the actual execution for better isolation?

Butch78:
Good tip, I made it target specific! 

On the use-case @davidhewitt : I embed the `monty` crate directly as a `wasm32-unknown-unknown` wasm-bindgen module for in-process, browser-side Python in a Spreadsheet cell. It runs synchronously on the eval thread with its own linear memory, and `max_duration` is the only limit that stops a non-allocating runaway like `while True: pass` (which never trips `max_allocations`/`max_memory`), hence needing the clock not to panic.

The results have been amazing, I'm actually lazy loading the binary and even with a bad connection It works nearly instantly 

I'm also run a second tier for the harder cases: I'm experimenting with `monty-cpython` in a container on Cloudflare (via `monty-pool`'s subprocess model) that browser sessions escalate to, so a user whose cell needs full CPython (numpy, pandas, unsupported stdlib) gets a real interpreter instead of hitting the in-browser sandbox's limits. I'm already tracking the 0.19 beta line (currently pinned at `0.0.19-beta.2`), so your subprocess / `monty-cpython` work is very much in my plan. The in-browser `monty` wasm stays the fast default, and anything it cannot run transparently falls through to CPython-in-container.

And yes, #525 looks very relevant. The `wasip1`-in-a-web-worker model would give us a real hard-kill (terminating the worker) and a working clock for free, which is compelling but the problem is Cloudfare don't really support wasip1 yet....

I am are currently using the in-process wasm-bindgen embedding, so moving to that shape is a larger change we are weighing; this fix unblocks the cooperative time budget on our current architecture in the meantime. 

I had a call with @samuelcolvin and he recommended focusing on browser first, I'm very close to that and would love to have call with you if wanted to see a demo @davidhewitt I'm hoping to soon have a beta out so people can try :)

davidhewitt (approved):
Makes sense, let's move forward with this 👍

PR #556, opened 2026-07-10
Add pytest-style assert failure messages

samuelcolvin:
Failed asserts now raise AssertionError with an introspected message so
sandboxed code and hosts can see the values involved instead of a blank
AssertionError:

- `assert 2 == 5` -> `AssertionError: assert 2 == 5` (all binary
  comparison operators; operand reprs truncated to 120 chars, evaluated
  exactly once via Dup2)
- `assert []` -> `AssertionError: assert []` (falsy value's repr for
  non-comparison tests)
- `assert 1 == 2, 'msg'` -> `AssertionError: msg\nassert 1 == 2`
  (explicit message first, detail appended on a new line; message
  expression still only evaluated on failure)

Implemented as four new terminator opcodes (AssertFailed[Cmp][Msg])
appended at the wire-format tail; comparison asserts reuse the existing
Compare* opcodes with the operands kept alive under a Dup2. If an
operand __repr__ or message __str__ raises, that part is dropped rather
than replacing the AssertionError.

This is a deliberate CPython divergence (documented in
limitations/assert.md), on by default on every surface. A new
compile-time flag (`MontyRun::new_with_options` +
`CompileOptions { assert_messages: false }`) restores CPython behavior;
the datatest CPython-parity harness uses it, so all fixtures are
unchanged.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01GgS4wonv73fps3T2GVUV5k

---
## Summary by cubic
Adds pytest‑style assert failure messages by default and fuses assert execution for faster passing asserts, with a streaming, byte‑capped operand repr (120 bytes by default) configurable from Rust, `monty-python`, and `monty-js`. Aligns float `%` with CPython and unifies comparison semantics so fused asserts and `Compare*` share one implementation.

- **New Features**
  - Bare `assert` failures include introspected detail; comparisons show both sides; other falsy tests show their repr; literal `False` stays plain `AssertionError`. Operand reprs stream through a byte‑capped writer (120 bytes default, cut on a character boundary); operands evaluate once; if `__repr__`/`__str__` fail, that part is dropped; traceback unchanged. `assert test, msg` yields `AssertionError: {msg}\nassert ...`; empty `''` is treated as absent; message stays lazy.
  - Assert opcodes collapse to a fused `Assert` (falls through on success) and a raising `AssertFailed`, with compare operator encoded in flags. Passing comparison asserts are one dispatch (~30% faster). Compare* and fused asserts share a single comparison implementation.
  - Float `%` now matches CPython (divisor sign, signed zero). `x % n == k` compiles as an ordinary comparison; the lhs shows the computed modulo. Zero‑division and `TypeError` messages match unfused `%`.

- **Migration**
  - Configure via `CompileOptions { assert_message_annotations: MaxBytes(n) | Off }` (default 120 bytes, `n` ≥ 1). The executor applies the byte limit at runtime; snapshots preserve it; dump format bumped to version 3. `MontyRun::new` and `MontyRepl::new` now take `CompileOptions`.
  - Bindings: `monty-python` `Monty.checkout(assert_message_annotations: bool | int)` and `monty-js` `checkout({ assertMessageAnnotations: boolean | number })`. `true`/absent = on with 120 bytes; `false`/`0` = off; integers ≥ 1 set a custom truncation (in bytes). Wire: `monty-proto` adds optional `Configure.assert_message_annotations` (`uint32`, bytes); `monty-cpython` ignores it.

<sup>Written for commit 5305e60a5f166d6f197dd9fe8e4dac24cd552a5a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/556?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/src/bytecode/vm/exceptions.rs
```diff
@@ -4,14 +4,20 @@ use super::VM;
...
     resource::ResourceTracker,
-    types::{PyTrait, Type},
+    types::{CmpOrder, PyTrait, Type},
     value::Value,
 };
 
+/// Max chars of a single operand's repr in an assert failure message —
+/// keeps `assert huge_list == other` messages readable while still showing
+/// enough of each value to diagnose the failure.
+const MAX_ASSERT_REPR_CHARS: usize = 120;
```
davidhewitt: pytest has `-v, -vv, -vvv` to render increasingly more diff, is that useful?

crates/monty/src/bytecode/vm/mod.rs
```diff
@@ -1623,6 +1624,22 @@ impl<'h, T: ResourceTracker> VM<'h, T> {
...
+                Opcode::AssertFailedMsg => {
+                    let error = self.assert_failed_msg(None);
+                    catch_sync!(self, cached_frame, error);
+                }
+                Opcode::AssertFailedCmpMsg => {
+                    let op = CmpOperator::from_operand(cached_frame.fetch_u8())
+                        .expect("invalid comparison operand in bytecode");
+                    let error = self.assert_failed_msg(Some(op));
+                    catch_sync!(self, cached_frame, error);
+                }
```
davidhewitt: There is some inconsistency here in grouping the second two opcodes but not the first ones:

- assert -> `assert_test()`
- assertcmp -> `assert_cmp(op)`
- assertfailedmsg -> `assert_failed_msg(None)`
- assertfailedcmpmsg -> `assert_failed_msg(Some(op))`

Should the first two be collapsed into an `Option<Op>` too? Should all four be collapsed into a two-arg function?

crates/monty/src/bytecode/op.rs
```diff
@@ -520,6 +520,22 @@ pub enum Opcode {
...
+    Assert,
+    /// Fused bare `assert lhs OP rhs`: stack [..., lhs, rhs] -> [...]. Operand
+    /// is `CmpOperator::as_operand`; comparison errors match `Compare*`.
+    AssertCmp,
+    /// Raise for failed `assert test, msg`: stack [..., test, msg] -> raises.
+    /// The message is first, with introspected detail appended when available.
+    AssertFailedMsg,
+    /// Raise for failed `assert lhs OP rhs, msg`: stack [..., lhs, rhs, msg] ->
+    /// raises. Operand is `CmpOperator::as_operand`.
+    AssertFailedCmpMsg,
```
davidhewitt: Should this opcode use flags so we can collapse to a single opcode? We have used half of our available opcode space, and this doesn't seem like it's a "hot" instruction, so reducing the amount of opcode space it consumes seems like the more correct tradeoff.

crates/monty/src/expressions.rs
```diff
@@ -827,19 +827,78 @@ pub enum Operator {
...
+    #[strum(serialize = "in")]
     In,
+    #[strum(serialize = "not in")]
     NotIn,
+    /// Fused `x % n == k` comparison (int-literal `k`), created by the prepare
+    /// phase as an optimization. Displays as `==` since it is an equality test
+    /// on the `%` result.
     // we should support floats too, either via a Number type, or ModEqInt and ModEqFloat
+    #[strum(serialize = "==")]
     ModEq(i64),
```
davidhewitt: Is this fused comparison really worth it? Seems like it is just a recipe for complexity and doesn't seem likely to be something which LLMs do so frequently the additional complexity is justified.

crates/monty/src/value.rs
```diff
@@ -584,27 +584,32 @@ impl<'h> PyTrait<'h> for Value {
...
             _ => Ok(None),
         }
     }
 
+    /// Fast-path `self % other == right_value` for the fused `CompareModEq` opcode.
+    ///
+    /// Returns `None` when this pair needs the slow path (`py_mod` + `py_eq`) —
+    /// non-numeric types, and any zero divisor so the fallback raises
+    /// `ZeroDivisionError` exactly like an unfused `%` would.
     fn py_mod_eq(&self, other: &Self, right_value: i64) -> Option<bool> {
```
davidhewitt: as above I'm sceptical this complexity is justified.

crates/monty/test_cases/chain_comparison__all.py
davidhewitt, line 1: I'm not totally convinced the churn on all the test files was worth it.

perf-plan.md
davidhewitt, line 1: I don't think this is intended to be here.

crates/monty/src/bytecode/op.rs
samuelcolvin: done.

crates/monty/src/bytecode/vm/mod.rs
samuelcolvin: done.

crates/monty/src/run.rs
```diff
@@ -201,6 +293,11 @@ pub(crate) struct Executor {
     /// One entry per input value, in the order the embedder passed them.
     /// Empty for the standard (non-REPL) execution path.
     pub(crate) input_slots: Vec<NamespaceId>,
+    /// Char-truncation limit for operand reprs in introspected assert failure
+    /// messages ([`CompileOptions::assert_message_annotations`]). Captured at
+    /// compile time but consumed at *runtime* — every VM constructed for this
+    /// program (fresh or snapshot-restored) receives it.
+    pub(crate) assert_repr_max_chars: u32,
```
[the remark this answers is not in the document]
samuelcolvin: that's okay.

crates/monty/src/bytecode/vm/exceptions.rs
samuelcolvin: fixed.

crates/monty/src/bytecode/vm/exceptions.rs
```diff
@@ -359,3 +484,81 @@ impl<T: ResourceTracker> VM<'_, T> {
...
     }
 }
+
+/// `repr()` of an assert operand, streamed into a [`TruncatingWriter`] capped
+/// at the compiled program's limit (default 120 bytes, `…` suffix). Aborting
+/// at the cap keeps pathological operands `O(cap)` — a fully materialized
+/// repr could blow the memory limit and turn a catchable `AssertionError`
+/// into a terminal `MemoryError`.
+fn assert_operand_repr(value: &Value, vm: &mut VM<'_, impl ResourceTracker>) -> RunResult<String> {
+    let mut writer = TruncatingWriter::new(vm.assert_repr_max_chars as usize);
```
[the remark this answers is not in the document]
samuelcolvin: Renamed the whole Rust surface to bytes: `AssertMessageAnnotations::MaxBytes`, `max_bytes()`/`from_max_bytes()`, `DEFAULT_MAX_BYTES`, and `vm.assert_repr_max_bytes` — the unit is UTF-8 bytes everywhere and every doc (proto, TS, Python, limitations) already said so.

limitations/assert.md
```diff
@@ -0,0 +1,75 @@
...
+  message-less `AssertionError`.
+
+## Formatting edge cases
+
+- Each operand's repr is truncated to 120 bytes (cut on a character boundary)
+  with a `…` suffix (configurable per session, see "Opt-out for embedders"
+  below).
+- A failing assert calls `repr()` on its operands, which CPython never does:
+  user `__repr__` side effects run and their cost counts against resource
+  limits. Rendering streams and stops at the truncation cap, so parts of a
```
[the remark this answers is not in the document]
samuelcolvin: Reworded to "Rendering is streamed and stops at the truncation cap".

PR #567, opened 2026-07-16
implement `NotImplemented` and use it for user equality

davidhewitt:
## Summary by cubic
Implements Python’s NotImplemented across the VM, type system, and JS/wire formats, and aligns equality with CPython: direct "==" preserves arbitrary __eq__ results with reflected and identity fallbacks; boolean comparisons and containers truth‑test equality; using NotImplemented in a boolean context raises TypeError.

- **New Features**
  - Added NotImplemented/NotImplementedType: correct repr/identity; type(NotImplemented) works; global name resolves to the singleton.
  - Wire/JS/proto: new tag 29 with round‑trip in JS, wire, and Python conversions.

- **Refactors**
  - Equality: direct "==" preserves the user value with reflected fallback; containers and membership accept identical objects before equality; "!=" negates equality; instances with __eq__ are unhashable.
  - Made truthiness fallible and propagated to VM ops (not, conditional jumps, fused asserts) and builtins (any, all, filter, zip(strict), json.dumps); boolean use of NotImplemented raises TypeError.

<sup>Written for commit b4c0bb37f26b7104db1cd8cc325778d19b6b9e5e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/567?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/src/types/instance.rs
```diff
@@ -349,6 +349,42 @@ pub(crate) fn instance_getattr(
...
+    let args = ArgValues::ArgsKargs {
+        args,
+        kwargs: KwargsValues::Empty,
+    };
+    let result = vm.evaluate_function("__eq__", func, args)?;
+    if result.is_not_implemented() {
+        result.drop_with(vm);
+        Ok(None)
+    } else {
+        let eq = result.py_bool(vm);
```
rewitt94: Here we force this into a bool - but I can do this in python (not sure why you would want to)

```
>>> class Expr:
...     def __eq__(self, other):
...         return "helloworld"
...
>>> Expr() == 1
'helloworld'
```

I don't think this is documented in limitations/divergences

crates/monty/src/types/instance.rs
```diff
@@ -349,6 +349,42 @@ pub(crate) fn instance_getattr(
...
+///
+/// `NotImplemented` is translated into the internal `None` sentinel so
+/// `Value::py_eq` can try reflected comparison or fall back to inequality.
+pub(crate) fn instance_eq(
+    self_id: HeapId,
+    other: &Value,
+    vm: &mut VM<'_, impl ResourceTracker>,
+) -> RunResult<Option<bool>> {
+    let class_id = instance_class(self_id, vm);
+    let Some(func) = class_member(class_id, "__eq__", vm) else {
```
rewitt94: There seem to be some issues around classes having a `__eq__` function but no `__hash__` function in monty.

crates/monty/src/value.rs
```diff
@@ -319,6 +324,7 @@ impl<'h> PyTrait<'h> for Value {
         match self {
             Self::Undefined => false,
             Self::Ellipsis => true,
+            Self::NotImplemented => true,
```
rewitt94: python won't let you call bool on NotImplemented 🤔

rewitt94 (changes requested):
Overall this looks cool, seems like a good thing to (not) implement 😅 

I've written some test cases files with claude, both regressions from the main branch and bugs/divergences (already on main) related to the change - please take a look - maybe we can add some to the repo?

 not-implemented-regression.py: pass (CPython), pass (main), FAIL (PR 567)
 not-implemented-bugs.py: pass (CPython), FAIL (main), FAIL (PR 567)

`not-implemented-bugs.py`

```
# Known divergences in class equality that Monty does not yet implement.
#
# Nothing here ever worked: each assert holds
# on CPython and fails on Monty regardless of whether `__eq__` is dispatched.
# They are the gaps left over once `__eq__` dispatch lands.


# === `==` dispatches __eq__ even for the same object ===
# Identity is not a shortcut for `==`: CPython calls `__eq__` and honours a
# `False` result. (The identity fast path only applies inside container lookups,
# which check `is` before `==` as an optimisation.)


class FalseEq:
    def __eq__(self, other):
        return False


c = FalseEq()
assert (c == c) is False, '`==` dispatches __eq__ rather than short-circuiting on identity'
assert (c != c) is True, '`!=` negates the dispatched __eq__ result for the same object'


# === Defining __eq__ without __hash__ makes instances unhashable ===
# Equal objects must hash equal, and a default identity hash cannot satisfy a
# custom `__eq__`, so CPython sets `__hash__ = None` rather than let the two
# disagree. Monty keeps the identity hash, silently missing dict/set lookups.


class EqNoHash:
    def __init__(self, x: int) -> None:
        self.x = x

    def __eq__(self, other):
        return isinstance(other, EqNoHash) and self.x == other.x


try:
    {EqNoHash(1): 'value'}
    assert False, 'expected an unhashable instance to be rejected as a dict key'
except TypeError as exc:
    assert str(exc) == "cannot use 'EqNoHash' as a dict key (unhashable type: 'EqNoHash')", str(exc)

try:
    {EqNoHash(1)}
    assert False, 'expected an unhashable instance to be rejected as a set element'
except TypeError as exc:
    assert str(exc) == "cannot use 'EqNoHash' as a set element (unhashable type: 'EqNoHash')", str(exc)

try:
    hash(EqNoHash(1))
    assert False, 'expected hash() of an unhashable instance to be rejected'
except TypeError as exc:
    assert str(exc) == "unhashable type: 'EqNoHash'", str(exc)


# === NotImplemented is not usable in a boolean context ===
# Returning NotImplemented from a comparison is easy to do by accident, so
# CPython refuses to treat it as a truth value rather than silently reading it
# as True.

try:
    bool(NotImplemented)
    assert False, 'expected bool(NotImplemented) to be rejected'
except TypeError as exc:
    assert str(exc) == 'NotImplemented should not be used in a boolean context', str(exc)
```

`not-implemented-regression.py`

```
# Equality dunders on user classes: `__eq__`, `__ne__` and their interaction
# with hashing.
#
# Every assert here holds on CPython *and* on identity-only equality, because
# each is written as an invariant rather than as a check of a specific dispatch
# mechanism: `!=` is never inferred from `==`, and a dict/set never disagrees
# with `==`. Dispatching `__eq__` without also dispatching `__ne__`/`__hash__`
# breaks both invariants, so these guard against a partial implementation.


# === __ne__ is dispatched rather than negating __eq__ ===
# A class may define both dunders and have them disagree, so the result of `!=`
# cannot be derived from `==`.


class BothDunders:
    def __eq__(self, other):
        return True

    def __ne__(self, other):
        return True


assert (BothDunders() != BothDunders()) is True, '`!=` dispatches __ne__ rather than negating __eq__'
assert (BothDunders() != 1) is True, '`!=` dispatches __ne__ when the other operand is not an instance'


class TruthyEq:
    def __eq__(self, other):
        return 'truthy'

    def __ne__(self, other):
        return True


assert (TruthyEq() != TruthyEq()) is True, '`!=` dispatches __ne__ whatever __eq__ returns'


# === Hashing agrees with equality ===
# Equal objects must hash equal, so a dict/set lookup must agree with `==`.
# These compare `==` against container membership rather than asserting a fixed
# result, so they hold whether equality is by identity or by `__eq__`.


class Key:
    def __init__(self, x: int) -> None:
        self.x = x

    def __eq__(self, other):
        return isinstance(other, Key) and self.x == other.x

    def __hash__(self):
        return hash(self.x)


a = Key(1)
b = Key(1)
d = {a: 'value'}
assert (a == b) == (b in d), 'dict membership agrees with =='
assert (a == b) == (d.get(b) == 'value'), 'dict lookup agrees with =='
assert (a == b) == (len({a, b}) == 1), 'set dedup agrees with =='
assert (a == b) == ([a].count(b) == 1), 'list.count agrees with =='
```

crates/monty/src/types/instance.rs
davidhewitt: I think let's document in limitations for now, this might be reasonable to change in the future but doesn't seem worth in this PR.

davidhewitt: Actually fixing some of the other cases fixed this at the same time.

crates/monty/src/types/instance.rs
davidhewitt: Added a mitigation of making types with `__eq__` unhashable like CPython pending support for user-defined `__hash__`.

crates/monty/test_cases/class__basic.py
```diff
@@ -81,6 +81,76 @@ def __init__(self) -> None:
...
+assert ep1 == ep2, 'custom __eq__ compares instance attributes'
+assert ep1 != ep3, 'custom __eq__ false result is respected'
+assert (ep1 == 1) is False, 'NotImplemented from instance equality falls back to unequal'
+assert (1 == ep1) is False, 'NotImplemented works when reflected equality reaches the instance'
+
+try:
+    {ep1: 'value'}
+    assert False, 'an instance with custom equality should be rejected as a dict key'
+except TypeError as exc:
+    assert str(exc) == "cannot use 'EqPoint' as a dict key (unhashable type: 'EqPoint')", 'dict key error'
```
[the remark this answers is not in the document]
davidhewitt: Monty matches Python 3.14 behavior

PR #568, opened 2026-07-16
split all binary operators to be proper methods on `PyTrait`

davidhewitt:
This is moving stuff out of `vm` code and into type-specific code.

Similarly there was lots of logic in `value.rs` for cross-type operations because the `PyTrait` signatures took `other: &Self` to e.g. `py_add`.

I split the `PyTrait` methods to have `PyTrait::py_add_impl` and `PyTrait::py_radd_impl` and allow for reflected operations the same way as CPython.

---
## Summary by cubic
Moves all binary operators to per-type `PyTrait` methods and unifies VM binary dispatch. Expands set-like dict view ops, tightens numeric and sequence behavior to match CPython, and validates memory use up front.

- New Features
  - `set`/`frozenset` and `dict_keys`/`dict_items` implement `|`, `&`, `^`, `-` (including reflected); unsupported ops (`+`, `*`, `<<`, `>>`) raise `TypeError` before hashing/materializing; unhashable element errors name both the element and the nested unhashable type.
  - `pow(base, exp, mod)` supports heap-backed ints and negative modulus; negative exponents are rejected; oversized non-modular results raise `OverflowError`.
  - Sequence `+` and `*` for `str`/`bytes`/`list`/`tuple` validate sizes up front; empty × huge count yields empty; large concatenations are rejected early under memory limits.
  - `bool` bitwise ops return `bool` when both sides are `bool`; mixing with `int` uses integer ops; shift counts (including heap-backed) respect memory limits.

- Refactors
  - All VM binary and in-place opcodes use one path; removed `BitwiseOp`/`SetBinaryOp`.
  - Added per-type `py_*_impl`/`py_r*_impl` on `PyTrait`; centralized long-int ops (`modular_pow`, `repeat_count`, `u128_into_value`); added `Set::preallocation_capacity` for dict-view ops; fast-pathed small integer operations.

<sup>Written for commit 06b012f33ca6e710b01960524b51e5f229973873. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/568?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/src/value.rs
```diff
@@ -2406,6 +2428,28 @@ impl BitwiseOp {
     }
 }
 
+/// Bitwise operation type retained while opcode call sites migrate to [`BinaryOp`].
+#[derive(Debug, Clone, Copy)]
+pub enum BitwiseOp {
```
rewitt94: This is almost not used outside `binary.rs` and `value.rs` - can we not achieve a full deletion/cutover in this PR?

crates/monty/src/value.rs
```diff
@@ -1967,83 +1986,70 @@ impl Value {
...
+    /// today it centralizes native `PyTrait` dispatch and final `TypeError`s.
+    pub fn py_binary(&self, other: &Self, op: BinaryOp, vm: &mut VM<'_, impl ResourceTracker>) -> RunResult<Self> {
+        if let Some(result) = self.py_binary_impl(other, op, vm)? {
+            Ok(result)
+        } else if let Some(result) = other.py_rbinary_impl(self, op, vm)? {
+            Ok(result)
+        } else {
+            let lhs_type = self.py_type(vm);
+            let lhs_name = self.py_type_name(vm);
+            Err(ExcType::binary_type_error(
```
rewitt94, line 1894: How this is called seems to be causing some regression of error messages - 

Claude wrote seem test cases to illustrate which on cpython (pass), main (pass), this branch (fail)

```
# Operators that dict views do not implement must report the operand types.
#
# Regression guard: a view must not be materialised into a set before the
# operator is checked. `dict_items` hashes its (key, value) pairs, so a view
# over unhashable values reports "unhashable" instead of the correct
# "unsupported operand" TypeError if the set is built too eagerly.

unhashable_values = {'a': [1]}

# === dict_items arithmetic operators ===
try:
    unhashable_values.items() + 1
    assert False, 'dict_items should not support +'
except TypeError as e:
    assert str(e) == "unsupported operand type(s) for +: 'dict_items' and 'int'", 'items + int names the operand types'

try:
    unhashable_values.items() * 2
    assert False, 'dict_items should not support *'
except TypeError as e:
    assert str(e) == "unsupported operand type(s) for *: 'dict_items' and 'int'", 'items * int names the operand types'

# === dict_items shift operators ===
try:
    unhashable_values.items() << 1
    assert False, 'dict_items should not support <<'
except TypeError as e:
    assert str(e) == "unsupported operand type(s) for <<: 'dict_items' and 'int'", 'items << int names the operand types'

try:
    unhashable_values.items() >> 1
    assert False, 'dict_items should not support >>'
except TypeError as e:
    assert str(e) == "unsupported operand type(s) for >>: 'dict_items' and 'int'", 'items >> int names the operand types'

# === Reflected forms take the same path ===
try:
    1 + unhashable_values.items()
    assert False, 'int + dict_items should not be supported'
except TypeError as e:
    assert str(e) == "unsupported operand type(s) for +: 'int' and 'dict_items'", 'reflected + names the operand types'

try:
    2 * unhashable_values.items()
    assert False, 'int * dict_items should not be supported'
except TypeError as e:
    assert str(e) == "unsupported operand type(s) for *: 'int' and 'dict_items'", 'reflected * names the operand types'

# === Supported set-like operators still build the set (and may raise unhashable) ===
# `-` is a real dict_items operator, so CPython materialises the view here too;
# the unhashable value is reported rather than an "unsupported operand" error.
try:
    unhashable_values.items() - [1]
    assert False, 'items difference over unhashable values should raise'
except TypeError:
    pass
```

rewitt94 (commented):
Not much to add from me - cool to the see the PyTrait in action!

I see the performance benchmarks are dropping - by 10% - but still microseconds (doesn't sound like an issue)? Curious if there is any variance on the benchmarks?

crates/monty/src/types/dict_view.rs
```diff
@@ -91,14 +91,48 @@ impl<'h> HeapRead<'h, DictKeysView> {
...
+        let mut result_guard = DropGuard::new(Set::with_capacity(capacity), vm);
+        let (result, vm) = result_guard.as_parts_mut();
+        let dict = self.dict(vm);
+        for key in other.iter() {
+            if dict.contains_key(key, vm)? {
+                result.add(key.clone_with_heap(vm), vm)?;
+            }
+        }
+        let (result, vm) = result_guard.into_parts();
+        Ok(Some(Value::Ref(vm.heap.allocate(HeapData::Set(result))?)))
```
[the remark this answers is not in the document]
davidhewitt: I have noted the issue with `heap.allocate` to think about for the future.

PR #570, opened 2026-07-17
HMAC signing  for `dump` / `restore`

samuelcolvin:
## Summary by cubic
Sign `dump()` bytes with HMAC-SHA256 and verify on restore to block forged or tampered dumps. Adds `dumpKey`/`dump_key` options (≥16 bytes), defaulting to a random per‑pool key; invalid dumps are rejected before contacting a worker.

- **New Features**
  - Host-side signing and verification for dumps (Rust); wasm path uses WebCrypto with the same byte format. Key import is lazy; caller-supplied key bytes are snapshotted; clear error when WebCrypto is unavailable in browsers.
  - JS (`@pydantic/monty`): `Monty.create({ dumpKey: string | Uint8Array })`; new `MontyInvalidDumpError`; worker transport signs/verifies; key is resolved lazily.
  - Python: `Monty(dump_key=...)` and `AsyncMonty(dump_key=...)` accept `str | bytes`; `ValueError` on mismatch/short keys; `AsyncMontyWebsocket` does no client-side signing.
  - Default is a random per-pool key; pass the same key to restore dumps across pools/processes.
  - On rejection, the session stays fresh and usable (no worker I/O happened).

- **Migration**
  - Old, unsigned dumps no longer load; take fresh dumps under the new scheme.
  - To share dumps across pools, set the same key (≥16 bytes).
  - Browser/wasm dump/load requires a secure context (WebCrypto); execution does not.
  - Expect `MontyInvalidDumpError` (JS) / `ValueError` (Python) on signature failure; sessions remain usable.
  - WebSocket transport: client does not sign/verify dumps; the server must handle signing.
  - Signing provides integrity only (dumps are not encrypted).

<sup>Written for commit 569bc72ab977d2169b262df09f42f9b0982cb444. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/570?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty-pool/src/dump_sign.rs
+++ b/crates/monty-pool/src/dump_sign.rs
@@ -43,7 +43,7 @@ const TAG_LEN: usize = 32;
 ///
 /// Supply one via [`crate::PoolConfig::dump_key`] to make dumps restorable
 /// across pools/processes; without one the pool generates a random ephemeral
-/// key at creation, so its dumps only restore into that same pool instance.
+/// key on first use, so its dumps only restore into that same pool instance.
 #[derive(Clone)]
 pub struct DumpKey(Vec<u8>);
 
@@ -59,12 +59,13 @@ impl DumpKey {
     }
 
     /// Generates a random 32-byte key from OS randomness, used when
-    /// [`crate::PoolConfig::dump_key`] is `None`.
-    pub(crate) fn ephemeral() -> Result<Self, PoolError> {
+    /// [`crate::PoolConfig::dump_key`] is `None`. Created lazily on the first
+    /// dump/restore, so pools that never dump draw no randomness. Panics if
+    /// the OS RNG fails — an unrecoverable environment fault, not a pool error.
+    pub(crate) fn ephemeral() -> Self {
         let mut key = vec![0u8; 32];
-        getrandom::fill(&mut key)
-            .map_err(|err| PoolError::Spawn(format!("failed to generate an ephemeral dump key: {err}")))?;
-        Ok(Self(key))
+        getrandom::fill(&mut key).expect("OS randomness unavailable — cannot generate a dump signing key");
+        Self(key)
     }
 
     /// Signs a worker dump envelope, prepending the version byte and MAC tag.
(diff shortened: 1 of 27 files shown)


crates/monty-pool/src/dump_sign.rs
```diff
@@ -0,0 +1,159 @@
...
+const TAG_LEN: usize = 32;
+
+/// How a pool signs [`crate::Checkout::dump`] bytes and verifies them in
+/// [`crate::Checkout::restore`]; set via [`crate::PoolConfig::dump_signing`].
+#[derive(Debug, Clone, Default)]
+pub enum DumpSigning {
+    /// A random key generated on the pool's first dump/restore: dumps only
+    /// restore into that same pool instance. The subprocess-transport default.
+    #[default]
+    Ephemeral,
```
davidhewitt, line 53: What's the advantage of having the default key be ephemeral? Seems like an easy user footgun as if they snapshot and then write to a file, then any subsequent later process can't read that file.

davidhewitt (commented):
I'm somewhat trying to work out the use case here.

It seems to me that this creates a fair amount of complexity to ultimately move the security model of any harness using monty to be "don't let users see the key" rather than "don't let users tamper with dumps".

I would imagine that most deployments wouldn't be letting users see the dumps anyway, so this doesn't add much for them except a ton of complexity. And for deployments that do let users see the dumps, they could add their own mac validation without it being part of monty.

I can somewhat see an advantage of signing keys which run on the server side of the proto so that even monty clients can't do tamper with the snapshot state. But I think if that is the case, a lot of the client logic doesn't need to exist and the proto definitions can just be clarified to say that the dump is not necessarily portable (e.g. a server could then implement signing or encryption of its dumps freely).

crates/monty-pool/src/dump_sign.rs
samuelcolvin: The idea was to be secure by design, but you might be right.

Instead of ephemeral keys, we could have the option have no signing at all.

PR #580, opened 2026-07-19
avoid collisions in `id`

davidhewitt:
`id` currently requires on hashing and mixing into 64 bits even though the state space is larger than 64 bits. maybe we can reduce the state space back down, but for now this will avoid the collision problem.

---
## Summary by cubic
Make `id()` collision-free by replacing hashing with a structural, injective identity encoding. `id()` now returns a real Python int (inline or heap `LongInt`), and function `repr` shows the exact 0x… from `hex(id(x))`.

- **Bug Fixes**
  - Removed 64-bit hashing/mixing that caused `id()` collisions.
  - External functions use name-based identities that are distinct across names and consistent across inline/heap forms.
  - `repr(function)` prints the same hex as `hex(id(function))`.

- **Refactors**
  - Added `identity` module: structural identities encode to Python ints; small ones are immediate, wide ones use heap `LongInt`.
  - `Value::is` compares full structural identities.
  - `id()` respects resource limits; wide identities may allocate.
  - Updated `py_repr_fmt` to accept `impl fmt::LowerHex` for IDs.
  - Tests: added allocation behavior for compact vs. wide `id()`; expanded external function tests to assert different names yield different `id()` values.

<sup>Written for commit 7103da343d14f75ac42e9712f990b8641eea7dd0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/580?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/builtins/id.rs
+++ b/crates/monty/src/builtins/id.rs
@@ -11,14 +11,5 @@ pub fn builtin_id(vm: &mut VM<'_, impl ResourceTracker>, args: ArgValues) -> Run
     let value = args.get_one_arg("id", vm.heap)?;
     defer_drop!(value, vm);
 
-    let id = value.id(vm);
-
-    // Python's id() returns a signed integer; reinterpret bits for large values
-    // On 64-bit: large addresses wrap to negative; on 32-bit: always fits positive
-    #[expect(
-        clippy::cast_possible_wrap,
-        reason = "Python id() returns signed; wrapping intentional"
-    )]
-    let id_i64 = id as i64;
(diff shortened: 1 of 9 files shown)


rewitt94:
Great, looks like a good bug to fix!

PR #581, opened 2026-07-19
Implement iter(callable, sentinel)

rewitt94:
Follows #579, and rebased onto #575.

The two-argument `iter()` form built a `TypeError: not yet supported`. It now
builds a `callable_iterator`: each step calls `callable()` with no arguments and
yields the result unless it `==` the sentinel, at which point iteration stops and
`done` latches so the iterator stays exhausted.

**Reimplemented on top of #575's `PyTrait` iterator protocol.** `callable_iterator`
is its own heap type alongside `list_iterator`, rather than a variant of
`IterValue`. That is what an earlier revision of this PR did, and it was the wrong
shape: it made `Callable` the largest `IterValue` variant, and since
`py_estimate_size` charges `size_of::<MontyIter>()` against the sandbox memory
budget, *every* iterator — `range`, lists, strings — grew by 8 bytes to pay for a
form only this one uses. As a heap type `MontyIter` is untouched, and
`type(iter(f, s)).__name__` is `'callable_iterator'` exactly as CPython reports it,
rather than the generic `'iterator'` that previously had to be documented as a
divergence.

A non-callable first argument is rejected when the iterator is built, as CPython
does, rather than at the first `next()`. That needs a callability predicate, so
`HeapData::is_callable` / `Value::is_callable` are added as the single source of
truth, kept in lockstep with `call_function` / `call_heap_callable` by
`debug_assert!`s in their "not callable" arms.

## Reference counting and GC

`callable` and `sentinel` are OWNED refs held in the iterator itself rather than in
a container it points at, so `py_dec_ref_ids` and `for_each_child_id` must both
enumerate them or the GC under-traces and frees a live object.

On exhaustion both refs are released, matching CPython's `calliter_iternext`, which
`Py_CLEAR`s them so a spent-but-still-referenced iterator does not pin the sentinel
or the callable's captured state.

**This touches `heap.rs`**, which the contributing guide flags as a critical safety
boundary — but only to add a `HeapData::CallableIterator` arm following the pattern
#575 established for `ListIterator`, rather than changing the semantics of the
existing `Iter` arm as the previous revision did.

Two unit tests cover the tracing, both verified to fail when that arm is stubbed
out: one pins the enumeration and its multiplicity when callable and sentinel alias
the same object (an id must appear twice, or trial deletion under-decrements), the
other proves a cycle *through* the callable is collected while its non-cycle
sentinel is released, exercising the mark and free phases together. The cycle is
built through the safe `HeapReader` API, so **neither test needs `unsafe`** — the
earlier revision reached into `entry.data.0.get()` and had to flag that for review.

## Re-entrancy

Because the callable re-enters Python, `py_next` resolves callable and sentinel to
owned clones and drops its heap borrow BEFORE the call. Holding it across a
re-entrant `next()` on the same iterator would alias the `UnsafeCell`.
`iter__reentrant.py` exercises exactly that and fails loudly under
`memory-model-checks` if the invariant regresses.

## Host boundary

`MontyType::CallableIterator` maps to CPython's `callable_iterator`, which — like
`list_iterator` — is not reachable as a `builtins` attribute and so is taken from a
throwaway two-argument `iter()`. Without that arm the mapping fell through to
`getattr(builtins, 'callable_iterator')`, so any `type(iter(f, s))` crossing to the
host raised `AttributeError`. `test_types.py` now round-trips the type in both
directions and fails with exactly that error if the arm is removed.

## Refactors

`OpaqueTarget::ListIterator` becomes `OpaqueTarget::Protocol`, covering any heap
iterator advanced through `py_next`, and dispatches via `HeapReadOutput::py_next`
instead of a per-type arm — so this type needed no new arm there, and neither will
the next one.

Arguments are parsed with `#[derive(FromArgs)]` (`style = unpack`, the
`PyArg_UnpackTuple` family `iter` belongs to) rather than a hand-rolled arity check
with manual `drop_with` on the error path.

## Divergences

`limitations/iter.md` records that the `callable` runs synchronously — one that
calls an external/OS function cannot suspend and raises `NotImplementedError`, the
same limitation as `map`/`filter`/`sorted(key=...)` — that a `StopIteration` raised
by the callable propagates where CPython treats it as clean exhaustion, and that the
sentinel comparison is `result == sentinel` where CPython does `sentinel == result`
(observable only with an asymmetric `__eq__`).

## Not in this PR

`monty -t` still rejects `iter` and `next` outright, because the trimmed
`builtins.pyi` declares neither. An earlier revision fixed that here, but the bug
predates this PR and applies equally to the one-argument form — the fix is identical
whether or not `iter(callable, sentinel)` exists — so it is split out into a
follow-up. @davidhewitt, that also moves your two review comments on the stub
declarations to that PR; the conclusion there is to take upstream typeshed verbatim,
since `SupportsNext`, `_SupportsNextT_co`, `_GetItemIterable` and `SupportsIter` are
all already vendored and nothing forced the divergence.

## Testing

`iter__sentinel.py` covers the sentinel protocol, `==`-by-value comparison, argument
errors and exhaustion behaviour; `iter__reentrant.py` the re-entrancy invariant;
`refcount__callable_iterator.py` proves the exhausted iterator released both owned
refs — the sentinel ends at 2, and was verified to be 3 when the release is removed.

`make lint-rs`, `make lint-py`, the 1060 test cases (plain and under
`ref-count-return`), the `iter` subset under `memory-model-checks`, the lib tests
and the 921 `pydantic_monty` tests all pass.


crates/monty/src/bytecode/vm/call.rs
```diff
@@ -494,6 +494,12 @@ impl<T: ResourceTracker> VM<'_, T> {
                 self.call_heap_callable(*heap_id, args)
             }
             _ => {
+                // Coupling check: reaching here means dispatch rejected the value,
+                // so `Value::is_callable` must agree it is not callable.
+                debug_assert!(
```
rewitt94, line 499: Are these assertions useful? Might be better do consolidate methods than link outcomes like this.

crates/monty-typeshed/vendor/typeshed/stdlib/builtins.pyi
```diff
@@ -1058,8 +1058,25 @@ else:
...
 
 def isinstance(obj: object, class_or_tuple: _ClassInfo, /) -> bool: ...
+
+# Diverges from upstream deliberately. Monty always returns its own generic
+# `iterator`, never the argument or a type-specific iterator, so the plain
+# `Iterator[_T]` return is more accurate here than upstream's
+# `_SupportsNextT_co`. Upstream's `_GetItemIterable` overload is omitted too:
+# Monty has no `__getitem__`-only iteration fallback.
+@overload
+def iter(object: Iterable[_T], /) -> Iterator[_T]: ...
```
davidhewitt: While true I imagine #575 would change this almost immediately, I don't think we need the divergence from typeshed here.

crates/monty-typeshed/vendor/typeshed/stdlib/builtins.pyi
```diff
@@ -1058,8 +1058,25 @@ else:
...
+@overload
+def iter(object: Callable[[], _T], sentinel: object, /) -> Iterator[_T]: ...
 def len(obj: Sized, /) -> int: ...
 
+# `Iterator` rather than upstream's `SupportsNext`: Monty's `next()` only
+# accepts a real iterator, and this composes with the `iter()` return above.
+@overload
+def next(i: Iterator[_T], /) -> _T: ...
+@overload
+def next(i: Iterator[_T], default: _VT, /) -> _T | _VT: ...
```
davidhewitt: Similar thinking as above.

crates/monty/src/types/iter.rs
```diff
@@ -48,16 +49,38 @@ impl MontyIter {
...
     /// - `iter(iterable)` - Returns an iterator for the iterable. If the argument is
     ///   already an iterator, returns the same object.
-    /// - `iter(callable, sentinel)` - Not yet supported.
+    /// - `iter(callable, sentinel)` - builds a `callable_iterator`.
     pub fn init(vm: &mut VM<'_, impl ResourceTracker>, args: ArgValues) -> RunResult<Value> {
-        let (iterable, sentinel) = args.get_one_two_args("iter", vm.heap)?;
+        // `iter` is positional-only — CPython's `PyArg_UnpackTuple` never sees
+        // keywords — so reject those first, with CPython's wording, then apply
+        // the same 1..=2 arity check `PyArg_UnpackTuple` does.
+        let mut positional = args.into_pos_only("iter", vm.heap)?;
```
davidhewitt: It might be nice to use `#[derive(FromArgs)]` to process the arguments here (not sure if it handles overloads?)

crates/monty/src/types/iter.rs
```diff
@@ -188,6 +243,24 @@ impl MontyIter {
...
+                if *done {
+                    return Ok(None);
+                }
+                // `self` is stack-owned here, so unlike `advance` this may keep
+                // `done` borrowed across the call.
+                let callable = callable.clone_with_heap(vm.heap);
+                let sentinel = sentinel.clone_with_heap(vm.heap);
+                let next = callable_next(callable, sentinel, vm)?;
+                if next.is_none() {
+                    *done = true;
```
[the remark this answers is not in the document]
davidhewitt: We could check what CPython does here. I don't think this is a huge issue either way as long as tests are consistent with implementation.

davidhewitt (approved):

crates/monty/src/types/iter.rs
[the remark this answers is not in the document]
rewitt94: Fixed — both refs are released when the sentinel is seen matching CPython's `Py_CLEAR`.

crates/monty-typeshed/vendor/typeshed/stdlib/builtins.pyi
rewitt94: Removed - doesn't need to be in this PR

crates/monty-typeshed/vendor/typeshed/stdlib/builtins.pyi
rewitt94: Removed - doesn't need to be in this PR

crates/monty/src/types/iter.rs
rewitt94: Done

crates/monty/src/args/bind_native.rs
```diff
@@ -120,15 +120,21 @@ fn bind_slow<const N: usize>(
...
+    // METH_VARARGS, so the call machinery rejects keywords BEFORE
+    // `PyArg_UnpackTuple` ever counts positionals — `iter(x=[1])` reports "takes
+    // no keyword arguments", not "expected at least 1 argument, got 0". Then
+    // `PyArg_UnpackTuple` checks its positional range; `at_most_total`
+    // reproduces the `PyArg_ParseTupleAndKeywords` total pre-count; the "at
+    // least M positional" check reproduces `_PyArg_UnpackKeywords` for C methods
+    // whose required positional-only params cannot be filled by keyword.
+    if matches!(spec.family, ErrorFamily::Unpack) {
+        if n_kw > 0 {
+            return Err(ExcType::type_error_no_kwargs(spec.func_name));
```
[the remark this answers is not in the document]
rewitt94: Dropped — main already fixed this in #584 via `kwarg_error_name`.

PR #582, opened 2026-07-19
Support class decorators

rewitt94:
Enables `@deco class Foo` in sandboxed Python. Decorators evaluate in the enclosing scope and apply bottom-up (`Foo = deco(Foo)`), matching CPython.

This adds **no new capability**: defining a class then reassigning it (`class Foo: ...; Foo = deco(Foo)`) was already possible — this is syntax sugar for it, previously rejected at parse time. Function and method decorators remain rejected.

### Changes

- **parse** — carry `decorator_list` on class defs instead of rejecting it, and locate the `class` statement at the `class` keyword rather than at the first decorator (see *Traceback locations* below).
- **prepare** — resolve decorator names in the enclosing scope, and walk decorator expressions in *all three* scope-analysis passes: reference collection, cell-var collection, and assigned-name collection.
- **compiler** — push decorator callables, build the class, then one `CallFunction 1` per decorator, applied bottom-up and each located at its own decorator expression. Decorators run as real bytecode, so they can suspend on external/OS calls.

### Scope analysis is load-bearing in three places

Decorators are arbitrary expressions evaluated in the *enclosing* scope, which makes them relevant to every scope-analysis pass. Missing any one is a silent miscompile, so all three are covered by regression tests:

1. **`collect_cell_vars_from_node`** — a nested scope inside a decorator expression that captures an enclosing local needs a cell. Without this, `MakeClosure` finds no cell reference on the stack at runtime:

   ```python
   def outer():
       n = 5
       @(lambda cls: setattr(cls, 'n', n) or cls)
       class C: ...
       return C
   ```

2. **`collect_referenced_names_from_node`** — decorator names resolve in the enclosing scope, so they are references of *that* scope.

3. **`collect_scope_info_from_node`** — a walrus in a decorator expression binds in the enclosing scope. Without this the binding escapes to module globals:

   ```python
   x = 'global'
   def f():
       @mk(x := 'local')
       class C: ...
       return x
   f(), x    # ('local', 'global') — previously ('local', 'local'), clobbering the global
   ```

### Traceback locations

Two source-location bugs, both fixed here, so tracebacks match CPython line-for-line:

1. **Apply site.** Each applying `CallFunction 1` was emitted with the whole statement's range, so every decorator in a stack reported the same location and nothing identified which one raised. Each call now carries its own decorator's range, iterated in reverse since decorators are pushed in source order and applied bottom-up.

2. **Class-body frame.** Ruff's `StmtClassDef::range` starts at the *first decorator*, where CPython locates the statement at the `class` keyword — so a decorated class whose **body** raised reported the `@deco` line and pulled the decorator into the displayed span. `class_keyword_range` scans back from the class name for the keyword; this is exact rather than heuristic, since only whitespace and line continuations may sit between `class` and the name. Undecorated classes take an early return, where ruff's range is already correct.

Two rendering divergences remain and are **not** decorator-specific: Monty draws flat spans where CPython narrows to the failing sub-expression (`~~~~~` vs `~~^~~`), and elides multi-line spans as `...<N lines>...`. Both reproduce on a plain multi-line call and are already documented in `limitations/exceptions.md`.

### Tests

`test_cases/decorator__class.py` — identity, class replacement, decorator factories, stacked bottom-up ordering, decorator-expression evaluation order (top-down, before the body), registry / method-injection / `__init__`-injection / `__repr__`-injection / singleton patterns, subscript and attribute expressions in decorator position, closure capture (single- and multi-hop), the cell-var and walrus regressions above, `__name__`/`__doc__` preservation, non-callable decorators, and clean unwinding when a decorator raises (looped, to catch refcount leaks).

Two traceback fixtures: `decorator__class_traceback.py` (a stack of three pins the middle one) and `decorator__class_body_traceback.py` (body raises under decorators — its decorator argument holds the literal text `class` and its header spacing is irregular, so the keyword scan cannot be fooled by either; it is ruff-format-excluded in `pyproject.toml` to preserve that spacing).

All dual-run against CPython 3.14. Full suite green under `--features memory-model-checks`. The obsolete `class_decorators_return_not_implemented_error` parse-error test is removed; the function- and method-decorator rejection tests remain.

### Docs

`limitations/classes.md` — class decorators listed as supported; function and method decorators still rejected. The general class-introspection limitation (`__dict__`, `__bases__`, `__annotations__`, `dir()` unavailable) is reworded to note its consequence for decorators: a decorator cannot discover the fields of the class it receives, so it can only act on names given to it up front.

---
## Summary by cubic
Adds class decorator support with CPython semantics. Decorators evaluate in the enclosing scope and apply bottom-up; function and method decorators remain rejected.

- **Bug Fixes**
  - Tracebacks: each decorator call pins its own expression; class-body frames start at the `class` keyword found from lexer tokens (robust to comments/strings).
  - Scope analysis walks decorator expressions in all passes (references, cell vars, assigned names/walrus) to avoid leaks and missing cells.

- **Refactors**
  - Build the parser once with `class` keyword offsets harvested from tokens, and add a free `code_range` helper for syntax errors; no behavior change.

<sup>Written for commit df45a6cf4e45a3b643d5311439ffa4ba689165fe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/parse.rs
+++ b/crates/monty/src/parse.rs
@@ -6,10 +6,11 @@ use ruff_python_ast::{
     self as ast, BoolOp, CmpOp, ConversionFlag as RuffConversionFlag, ElifElseClause, Expr as AstExpr,
     InterpolatedStringElement, Keyword, Number, Operator as AstOperator, ParameterWithDefault, Stmt, UnaryOp,
     name::Name,
+    token::TokenKind,
     visitor::{Visitor, walk_expr},
 };
 use ruff_python_parser::parse_module;
-use ruff_text_size::{Ranged, TextRange};
+use ruff_text_size::{Ranged, TextRange, TextSize};
 
 use crate::{
     StackFrame,
@@ -165,6 +166,15 @@ pub(crate) fn parse_with_interner(
     let mut parser = Parser::new(code, filename, interner);
     let parsed =
         parse_module(code).map_err(|e| ParseError::syntax(e.error.to_string(), parser.convert_range(e.range())))?;
+    // Harvested before `into_syntax` drops the token stream: `class_keyword_range`
+    // needs keyword offsets, and the lexer has already ruled out `class` occurring
+    // inside a comment or string literal.
+    parser.class_keyword_offsets = parsed
+        .tokens()
+        .iter()
+        .filter(|token| token.kind() == TokenKind::Class)
+        .map(Ranged::start)
+        .collect();
     let module = parsed.into_syntax();
     let nodes = parser.parse_statements(module.body)?;
     Ok(ParseResult {
@@ -187,6 +197,11 @@ pub struct Parser<'a> {
     /// Starts at MAX_NESTING_DEPTH and decrements on each nested level.
     /// When it reaches zero, we return a "Source is too deeply nested" syntax error.
     depth_remaining: u16,
+    /// Source offset of every `class` keyword token, ascending. Populated by
+    /// [`parse_with_interner`] straight after parsing (empty until then, which is
+    /// why nothing may read it during construction) and consumed only by
+    /// [`Parser::class_keyword_range`].
+    class_keyword_offsets: Vec<TextSize>,
 }
 
 impl<'a> Parser<'a> {
@@ -197,6 +212,7 @@ impl<'a> Parser<'a> {
             filename_id,
             interner,
             depth_remaining: MAX_NESTING_DEPTH,
+            class_keyword_offsets: Vec::new(),
         }
     }
 
@@ -834,20 +850,27 @@ impl<'a> Parser<'a> {
     /// The range of a `class` statement from the `class` keyword, excluding
     /// decorators: ruff's `StmtClassDef::range` starts at the first decorator
     /// where CPython starts at the keyword, which would otherwise show decorator
-    /// lines in a class-body traceback frame. Scanning back for the keyword is
-    /// exact — only whitespace and continuations may sit between it and the name.
+    /// lines in a class-body traceback frame.
+    ///
+    /// The keyword is located from the lexer's tokens rather than by searching the
+    /// source text, so a `class` inside a comment or a decorator's string argument
+    /// cannot be mistaken for it. Offsets are ascending, so this class's keyword is
+    /// simply the first one at or after its final decorator.
     fn class_keyword_range(&self, class: &ast::StmtClassDef) -> CodeRange {
-        let start = if class.decorator_list.is_empty() {
+        let start = match class.decorator_list.last() {
             // Undecorated: ruff's range already starts at the keyword.
-            class.range.start().into()
-        } else {
-            let name_start = usize::from(class.name.range.start());
-            // Always fits in `u32`; the fallback is unreachable (a class statement
-            // always contains the keyword) but avoids a panic path.
-            self.code[..name_start]
-                .rfind("class")
-                .and_then(|i| u32::try_from(i).ok())
-                .unwrap_or_else(|| class.range.start().into())
+            None => class.range.start().into(),
+            Some(last_decorator) => {
+                let after_decorators = last_decorator.range.end();
+                let index = self.class_keyword_offsets.partition_point(|&o| o < after_decorators);
+                // Bounded by the name so a malformed lookup cannot borrow a later
+                // class's keyword; the fallback is unreachable (a decorated class
+                // always has a keyword in this window) but avoids a panic path.
+                self.class_keyword_offsets
+                    .get(index)
+                    .filter(|&&offset| offset < class.name.range.start())
+                    .map_or_else(|| class.range.start().into(), |&offset| offset.into())
+            }
         };
         CodeRange {
             filename: self.filename_id,
--- a/crates/monty/src/parse.rs
+++ b/crates/monty/src/parse.rs
@@ -679,16 +679,19 @@ impl<'a> Parser<'a> {
     /// expressions (`name = <expr>` / `name: T = <expr>`). Every member name is
     /// recorded in `members`, in source order, for namespace assembly.
     ///
-    /// `pass` and `...` are ignored; a leading docstring becomes a synthesized
-    /// `__doc__` member (defaulting to `None`). Inheritance/metaclass syntax
-    /// (`class Foo(Bar):`), class/method decorators, and anything else in the
-    /// body (arbitrary control flow, complex targets) are rejected with a
-    /// not-implemented error, reserving the syntax for later.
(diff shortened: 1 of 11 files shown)


limitations/language.md
```diff
@@ -13,11 +13,16 @@ any code runs.
...
+  evaluated in the enclosing scope and applied bottom-up. Rejected at parse time
+  on functions and methods, so `@classmethod`, `@staticmethod`, `@property` and
+  any decorator on a `def` are unavailable. Previously top-level `def` decorators
   were silently ignored; they now raise `NotImplementedError` rather than
-  changing behaviour without warning.
+  changing behaviour without warning. Stdlib class decorators are unavailable
+  too, but because the modules providing them are missing rather than for any
+  decorator-specific reason — `@dataclass`, `@total_ordering` and `@unique` all
+  fail at the `import`, since there is no `dataclasses`, `functools` or `enum`
+  module. See [classes.md](classes.md).
```
davidhewitt: I suggest binning everything in this bullet which isn't directly relevant to decorators as they function now.

```suggestion
- **Decorators** (`@deco`) — supported on classes, taking any callable in scope,
  evaluated in the enclosing scope and applied bottom-up. Rejected at parse time
  on functions and methods, so `@classmethod`, `@staticmethod`, `@property` and
  any decorator on a `def` are unavailable.
```

limitations/classes.md
```diff
@@ -177,8 +178,14 @@ e.g. return a `dict` of the fields.
...
-- `@classmethod`, `@staticmethod`, `@property`, and any other class/method
-  decorators (rejected at parse time).
+- Function and method decorators — `@classmethod`, `@staticmethod`, `@property`,
+  and any decorator on a top-level `def` or a method (rejected at parse time).
+  Class decorators are supported.
+- **Classes are barely introspectable**: `__dict__`, `__bases__`,
+  `__annotations__` and `dir()` are all unavailable (`cls.__name__` works). This
+  applies to every class, not just decorated ones, but it is what stops a class
+  decorator discovering the fields of the class it receives — it can only act on
+  names it was given up front.
```
davidhewitt: I think decorators are only one of the things which lose out from missing introspection, it's an unrelated detail really.

```suggestion
- **Classes are barely introspectable**: `__dict__`, `__bases__`,
  `__annotations__` and `dir()` are all unavailable (`cls.__name__` works).
```

limitations/classes.md
```diff
@@ -34,7 +34,8 @@ divergences below apply to. Working, CPython-matching features: instance
 methods, `__init__` (full parameter shapes), instance and class attribute
 get/set (including `setattr(Foo, ...)` and function-attributes-become-methods),
 bound methods, class variables (arbitrary expressions, evaluated in a real
-suspendable class-body scope), `__repr__`/`__str__`/`__enter__`/`__exit__`
+suspendable class-body scope), **class decorators** (`@deco class Foo`, applied
+bottom-up), `__repr__`/`__str__`/`__enter__`/`__exit__`
```
davidhewitt:
```suggestion
suspendable class-body scope), **class decorators** (`@deco class Foo`), `__repr__`/`__str__`/`__enter__`/`__exit__`
```

crates/monty/test_cases/decorator__class_body_traceback.py
```diff
@@ -0,0 +1,29 @@
+# A traceback from a *decorated* class body locates the frame at the `class`
+# keyword, not the first decorator. The decorator argument holds the literal
+# text `class` and the header spacing is irregular: neither may fool the scan.
```
davidhewitt: What is "the scan"?

crates/monty/test_cases/decorator__class.py
```diff
@@ -0,0 +1,438 @@
...
+
+    return middle()
+
+
+assert grandparent().mark == 1
+
+
+# Regression: a *nested scope inside the decorator expression* that captures an
+# enclosing local needs a cell var too. Scope analysis must walk the decorator
+# expressions, not just the class body, or MakeClosure finds no cell at runtime.
```
davidhewitt: Let's kill the implementation detail hint

```suggestion
# enclosing local needs a cell var too.
```

crates/monty/src/parse.rs
```diff
@@ -823,10 +826,36 @@ impl<'a> Parser<'a> {
...
+            // Undecorated: ruff's range already starts at the keyword.
+            class.range.start().into()
+        } else {
+            let name_start = usize::from(class.name.range.start());
+            // Always fits in `u32`; the fallback is unreachable (a class statement
+            // always contains the keyword) but avoids a panic path.
+            self.code[..name_start]
+                .rfind("class")
+                .and_then(|i| u32::try_from(i).ok())
+                .unwrap_or_else(|| class.range.start().into())
```
davidhewitt: Why the `rfind`? Seems like it might break on nested classes. Can it instead use the decorator list and start with a scan beginning at the end of the decorator list?

crates/monty/test_cases/decorator__class_body_traceback.py
rewitt94: The `rfind` method you commented on

crates/monty/src/parse.rs
```diff
@@ -823,10 +862,43 @@ impl<'a> Parser<'a> {
...
+    /// The range of a `class` statement from the `class` keyword, excluding
+    /// decorators: ruff's `StmtClassDef::range` starts at the first decorator
+    /// where CPython starts at the keyword, which would otherwise show decorator
+    /// lines in a class-body traceback frame.
+    ///
+    /// The keyword is located from the lexer's tokens rather than by searching the
+    /// source text, so a `class` inside a comment or a decorator's string argument
+    /// cannot be mistaken for it. Offsets are ascending, so this class's keyword is
+    /// simply the first one at or after its final decorator.
+    fn class_keyword_range(&self, class: &ast::StmtClassDef) -> CodeRange {
```
davidhewitt, line 879: Do you think it would be reasonable for `class.range` to resolve to start at the `class` keyword, or otherwise ask ruff to preserve sufficient information in the AST? Maybe we can open an upstream issue and hope to simplify here in the future.

davidhewitt (approved):

crates/monty/src/parse.rs
rewitt94: I can try and find some context of why ruff doesn't preserve this in the AST

PR #588, opened 2026-07-20
migrate iterator callsites to `py_iter`

davidhewitt:
followup to #575 

---
## Summary by cubic
Migrate all iterator call sites to the Python iteration protocol (`py_iter`/`py_next`), replacing direct `MontyIter` usage to simplify ownership, unify timeout checks, and keep native allocations within the resource tracker.

- **Refactors**
  - Replaced `MontyIter::new`/`for_next` with `Value::py_iter`/`into_py_iter` + `py_next` across builtins (`any`, `enumerate`, `filter`, `map`, `min/max`, `reversed`, `sorted`, `sum`, `zip`, `str.join`, `bytes.join`) and core types (`list`, `tuple`, `set`, `dict`, `str`, `bytes`, `value`); multi-iter loops (e.g., `zip`) use retained iterator reads.
  - Added `iter::collect_owned_iterable` and routed collectors through retained Python iterators for incremental budget checks; used by `reversed`, `sorted`, `list`, `tuple`, and `list.extend`.
  - Switched capacity reservations to `iter::checked_preallocation_hint` using iterator size hints (e.g., `map`, `set`).
  - Removed `MontyIter::for_next`, `preallocation_hint`, and `.collect()`; iteration now advances via `py_next` on a retained iterator value.
  - Introduced `Value::into_py_iter` to enter the iteration protocol while releasing the original reference.

- **Migration**
  - New pattern: `let it = value.into_py_iter(vm)?; defer_drop!(it, vm); let mut it = it.read(vm); while let Some(x) = it.py_next(vm)? { ... }`

<sup>Written for commit 22882981c36c2998a9d607d05c74a9febcafcfd1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/588?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #589, opened 2026-07-20
Allow any iterable after `*`, not just five container types

rewitt94:
## The issue

```python
[*range(3)]     # TypeError: Value after * must be an iterable, not range
list(range(3))  # [0, 1, 2]
```

The unpacking sites read items straight out of a container's backing storage, so they matched on concrete types — `List`/`Tuple`/`Set`/`Dict`/`Str`. Anything without an arm was rejected as "not an iterable", which was untrue: `range`, `bytes`, `frozenset`, named tuples and the three dict views are all iterable, and `MontyIter` has always handled them.

The list recorded what someone had written a branch for rather than anything about iterability — `Counter` works only because it happens to be stored as a `Dict`. It also grew steadily, gaining `Iter` and then `ListIterator`.

Affected today, all six unpacking forms, every one of which works under `list()`:

| | before | after |
|---|---|---|
| `[*range(3)]` | `TypeError` | `[0, 1, 2]` |
| `[*b'ab']` | `TypeError` | `[97, 98]` |
| `[*frozenset([1])]` | `TypeError` | `[1]` |
| `[*d.keys()]` / `.values()` / `.items()` | `TypeError` | works |
| `[*sys.version_info]` | `TypeError` | works |
| `a, b = range(2)` | `TypeError` | works |

## The fix

Each of the four sites (`list_extend`, `set_extend`, `unpack_sequence`, `unpack_ex`) drains its operand through the iteration protocol. `collections.rs` no longer names a single container type: every site is now an iterability check, for its own error message, followed by a collect. `allocate_char` is not even imported there any more.

Iterability is answered by a new `py_is_iterable` on `PyTrait`, mirroring `py_is_context_manager`: each type declares its own, and `HeapReadOutput` dispatches exhaustively like `py_iter` already does. A new heap variant fails to compile until it says which it is.

## Why not just try to iterate and map the failure

Each site reports its own CPython-specific message, so it must know before it starts:

| | message |
|---|---|
| `[*x]` | `Value after * must be an iterable, not int` |
| `{*x}` | `'int' object is not iterable` |
| `a, b = x` | `cannot unpack non-iterable int object` |

## The one fact that really is type-specific

`too many values to unpack` carries a total for some sources and not others, and the rule is CPython's, not ours: only an **exact** `list`, `tuple` or `dict` reports one. Those three are unpacked without the iterator protocol, so their length is known; everything else — a subclass of those same three, `str`, `bytes`, `set`, `frozenset`, a dict view, a namedtuple, `range`, any iterator — stops at the first surplus item and never learns it.

So the sites keep the bounded collect, and the length moves to the types as a declarative `py_unpack_total` on `PyTrait`, sitting beside `py_is_iterable`. It is consulted only when wording that error, never to decide how to iterate.

Monty matched neither edge of this rule before, in both directions:

| | before | after / CPython |
|---|---|---|
| `a, b = 'abc'` | `too many values to unpack (expected 2, got 3)` | `too many values to unpack (expected 2)` |
| `a, b = {1: 'a', 2: 'b', 3: 'c'}` | `too many values to unpack (expected 2)` | `too many values to unpack (expected 2, got 3)` |

Both were pre-existing, and both fell out of removing the fast paths — `str` had one and so reported a total it should not have, `dict` had none and so omitted one it should have had. All eleven source types are now asserted against CPython. *Too few* values always carries the total, whatever the source: the iterable has been drained by then.

## Docs

`limitations/language.md` loses the claim that starred expressions are rejected at parse time (`[*xs, *ys]` and `f(*args)` have both always worked), and gains the one remaining divergence: `f(*non_iterable)` reports the list-literal message, where CPython names the callable by its module-qualified `__qualname__` (`__main__.f()`, `__main__.outer.<locals>.inner()`). Matching that needs function `__qualname__` and module-qualified names, both repo-wide gaps — fixing only the local part would swap one wrong message for another that is wrong in more cases.

---
## Summary by cubic
Allow any iterable after `*` in list/set/tuple literals, function calls, and assignment unpacking by delegating to the iteration protocol. Matches `list()` behavior and CPython error messages, including “too many values to unpack”.

- **Bug Fixes**
  - All unpacking sites first check `py_is_iterable` for their own error, then collect via `py_iter`/`py_next` — no per-type fast paths.
  - “Too many values to unpack” now reports a total only for exact `list`/`tuple`/`dict`, special-cased locally in `unpack_sequence`; `py_unpack_total` removed.
  - Implemented `py_is_iterable` for iterable types (interned/heap `str`/`bytes`, dict views, `range`, `frozenset`, named tuples, `CallableIterator`, iterators) with exhaustive dispatch in `HeapReadOutput` and coverage for interned bytes in `Value`.
  - Tests expanded to cover all six unpacking forms, including `set` at call/assignment sites; fixed Python lint by typing `bad_values`. Docs note starred-unpacking support and the one call-site message divergence.

<sup>Written for commit 83f7f8ae1d4f56880f0de39c067e8219f78a739b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/589?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/src/bytecode/vm/collections.rs
davidhewitt, line 1: it would be nice to move type-specific knowledge out of this file, do you think it's possible to use `py_iter()` and `py_next()` in similar style to #588 ?

crates/monty/src/bytecode/vm/collections.rs
```diff
@@ -108,42 +108,14 @@ impl<T: ResourceTracker> VM<'_, T> {
...
-                items
-            }
-            _ => {
-                let type_ = iterable.py_type_name(this);
-                return Err(ExcType::type_error_value_after_star(&type_));
-            }
-        };
+        // Every iterable is drained through the iteration protocol, so this file
+        // holds no per-type knowledge; `py_is_iterable` is asked first only
+        // because the message below is this site's own.
```
davidhewitt: I think comment here is unnecessary

```suggestion
```

crates/monty/src/bytecode/vm/collections.rs
```diff
@@ -520,100 +462,44 @@ impl<T: ResourceTracker> VM<'_, T> {
...
+            return Err(unpack_type_error(&type_name));
+        }
+        // Asked before iterating: a `list`/`tuple`/`dict` reports its length in
+        // the "too many" message, and reading it afterwards would be too late
+        // for a source that iteration mutates.
+        let total = value.py_unpack_total(this);
+        // Pull one past `count` so a too-long iterable is detected without
+        // draining it — CPython stops consuming there too, which is why every
+        // other type has no total to report.
+        let items = collect_iterable_bounded(value, count + 1, this)?;
```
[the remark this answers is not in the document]
davidhewitt: I don't think we need the special case for performance.

crates/monty/src/types/py_trait.rs
```diff
@@ -510,6 +510,40 @@ pub(crate) trait PyTrait<'h> {
...
+    /// sites that must report their *own* error for a non-iterable — the `*`
+    /// literal unpack (`Value after * must be an iterable`), set unpack
+    /// (`'x' object is not iterable`) and assignment unpacking (`cannot unpack
+    /// non-iterable x object`) all word it differently, so they have to ask
+    /// before delegating rather than map a failure afterwards.
+    ///
+    /// Mirrors [`py_is_context_manager`](PyTrait::py_is_context_manager): a type
+    /// that can be iterated MUST return `true` here, or it will be reported as
+    /// non-iterable by those sites while `list()` and `for` still accept it.
+    fn py_is_iterable(&self, _vm: &VM<'h, impl ResourceTracker>) -> bool {
```
davidhewitt, line 526: I went back and forth on whether this adds value over just calling `py_iter` and seeing if it fails; let's keep this for now but might revise later.

crates/monty/src/bytecode/vm/collections.rs
```diff
@@ -520,100 +462,44 @@ impl<T: ResourceTracker> VM<'_, T> {
...
-                        return Err(unpack_type_error(&type_name));
-                    }
+        if !value.py_is_iterable(this) {
+            let type_name = value.py_type_name(this);
+            return Err(unpack_type_error(&type_name));
+        }
+        // Asked before iterating: a `list`/`tuple`/`dict` reports its length in
+        // the "too many" message, and reading it afterwards would be too late
+        // for a source that iteration mutates.
+        let total = value.py_unpack_total(this);
```
davidhewitt: It looks like cpython special-cases this to dict, tuple, list:

https://github.com/python/cpython/blob/b3e046cdb4e9ac869b77022579d47ddb848c4fda/Python/ceval.c#L2385-L2394

so in contrast to the other "let's make the rest of the code generic using pytrait" I think let's not add `py_unpack_total` to `PyTrait` and special-case the exception messages here.

davidhewitt (commented):
Generally looks good, some small nits, thanks

crates/monty/src/bytecode/vm/collections.rs
rewitt94: Yup agreed!

crates/monty/src/bytecode/vm/collections.rs
rewitt94: Done

crates/monty/test_cases/star__unpack_iterables.py
```diff
@@ -0,0 +1,388 @@
...
+assert (hb3, hbrest) == (97, [98])
+
+# === The five types that used to have a fast path, at every site ===
+# list/tuple/set/dict/str were read straight out of their backing storage by
+# each unpacking site; they now go through the iteration protocol like anything
+# else. These pin the items themselves, not just that iteration was accepted —
+# the rest of this file leans on types that never had a fast path to lose.
+lst = [1, 2]
+tup = (1, 2)
+st = {1, 2}
```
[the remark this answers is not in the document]
rewitt94: Done

davidhewitt (approved):
Nice!

PR #593, opened 2026-07-21
Populate class __annotations__ (stringized)

rewitt94:
A class body's `x: T` annotations populate `__annotations__` as an ordered dict of stringized annotations.

**Stringized is interim.** PEP 563 never became CPython's default — PEP 649 did, in 3.14. The blocker isn't the annotation machinery, it's that Monty has no generic types: `list[int]` and `int | None` raise `TypeError`, so evaluated annotations would fail on the most common forms. Runtime `types.GenericAlias` is the prerequisite. Values are documented as provisional; keys and their order are stable either way — which is all a class transformer needs. CPython's own `@dataclass` does read the values, as evaluated objects, but keeps a string path for `ClassVar`/`InitVar` so PEP 563 code still works — so stringized annotations are enough to build on.

**Matching CPython.** Annotations are stringized by *unparsing* the expression, as CPython's stringizer does, rather than slicing source text — so spacing, line breaks and quote style don't leak into the value:

```python
class C:
    a: list [ int ]     # 'list[int]'
    b: dict[str,int]    # 'dict[str, int]'
    c: "int"            # "'int'"
```

**`__future__` imports** are handled as the compiler directives they are: honoured features are no-ops, in the absolute form only. Anything that would bind a name the no-op doesn't create is rejected rather than silently ignored — `barry_as_FLUFL` (PEP 401), and aliases such as `import annotations as ann`.

`test_cases/class__annotations.py` dual-runs green against CPython 3.14 and includes a worked `@mini_dataclass` that discovers its fields from `__annotations__`. Divergences are documented in `limitations/{typing,language,classes}.md`.

---
## Summary by cubic
Populates class `__annotations__` with ordered, CPython‑normalized, stringized annotations, and treats absolute `from __future__ import ...` as a no‑op. Stringization rebuilds string, bytes, and f-string literals canonically from their value to match CPython.

- **New Features**
  - Class annotations: build `__annotations__` in source order from `name: T`; values are strings from AST unparse with canonical literal normalization (merge adjacent string and f-string parts; fold raw/triple-quoted forms; prefer single quotes; keep the `u` prefix; apply to `bytes` literals too).
  - Behavior: `name: T = v` also creates a class variable; bare `name: T` does not bind a value; complex targets with a value are rejected; bare complex annotations are ignored; assigning `__annotations__` in a class body that also has annotated names is rejected; an empty dict is synthesized unless the body binds `__annotations__` itself.
  - `__future__`: absolute `from __future__ import x` is a no-op for supported features (including `annotations`); unknown features raise `SyntaxError`; `barry_as_FLUFL` and aliasing (`as ...`) raise `NotImplementedError`; relative `from .__future__ import ...` is treated as a normal import and errors accordingly.

- **Dependencies**
  - Add `ruff_python_codegen` and `ruff_source_file` to unparse and normalize annotations consistently.

<sup>Written for commit 950aa15a9c6252a82ed54516f520b4dcf082b250. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/593?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/stringize.rs
+++ b/crates/monty/src/stringize.rs
@@ -10,7 +10,7 @@
 use ruff_python_ast::{
     self as ast, AtomicNodeIndex, Expr as AstExpr,
     str::{Quote, TripleQuotes},
-    str_prefix::StringLiteralPrefix,
+    str_prefix::{ByteStringPrefix, StringLiteralPrefix},
     visitor::transformer::{Transformer, walk_expr, walk_f_string},
 };
 use ruff_python_codegen::{Generator, Indentation};
@@ -38,6 +38,7 @@ pub(crate) fn stringize_annotation(annotation: &mut AstExpr) -> String {
 /// | `f"x" "y"`     | `f'x' 'y'`     | `f'xy'`             |
 /// | `r"raw\d"`     | `r'raw\d'`     | `'raw\\d'`          |
 /// | `"""triple"""` | `"""triple"""` | `'triple'`          |
+/// | `b"foo" b"bar"`| `b"foo" b"bar"`| `b'foobar'`         |
 struct CanonicalStringLiterals;
 
 impl Transformer for CanonicalStringLiterals {
@@ -46,6 +47,7 @@ impl Transformer for CanonicalStringLiterals {
     fn visit_expr(&self, expr: &mut AstExpr) {
         match expr {
             AstExpr::StringLiteral(s) => rebuild_string_literal(s),
+            AstExpr::BytesLiteral(b) => rebuild_bytes_literal(b),
             AstExpr::FString(f) if f.value.is_implicit_concatenated() => merge_f_string_parts(f),
             _ => {}
         }
@@ -96,6 +98,24 @@ fn rebuild_string_literal(expr: &mut ast::ExprStringLiteral) {
         flags: canonical_string_flags(expr.value.first_literal_flags()),
     };
     expr.value = ast::StringLiteralValue::single(canonical);
+}
+
+/// The `bytes` counterpart of [`rebuild_string_literal`], collapsing
+/// `b"foo" b"bar"` into `b'foobar'`.
+///
+/// Simpler than the `str` case only in the flags: `bytes` has no `u` prefix, so
+/// the canonical form keeps no prefix at all.
+fn rebuild_bytes_literal(expr: &mut ast::ExprBytesLiteral) {
+    let canonical = ast::BytesLiteral {
+        range: expr.range,
+        node_index: AtomicNodeIndex::default(),
+        value: expr.value.bytes().collect(),
+        flags: ast::BytesLiteralFlags::empty()
+            .with_prefix(ByteStringPrefix::Regular)
+            .with_quote_style(Quote::Single)
+            .with_triple_quotes(TripleQuotes::No),
+    };
+    expr.value = ast::BytesLiteralValue::single(canonical);
 }
 
 /// Collapses `f"x" "y"` into the single f-string `f'xy'`.
--- a/crates/monty/src/stringize.rs
+++ b/crates/monty/src/stringize.rs
@@ -57,29 +57,29 @@ pub(crate) fn stringize_annotation(annotation: &mut AstExpr) -> String {
 struct CanonicalStringLiterals;
 
 impl Transformer for CanonicalStringLiterals {
-    fn visit_string_literal(&self, string_literal: &mut ast::StringLiteral) {
-        string_literal.flags = canonical_string_flags(string_literal.flags);
+    // Rebuilding is unconditional rather than reserved for the spellings that
+    // happen to diverge: the value is the truth and the spelling is noise, so
+    // there is no "unaffected" case to preserve. One path, not a rule plus
+    // exceptions — which is also why the whole rewrite lives here rather than
+    // being split with `visit_string_literal`.
+    fn visit_expr(&self, expr: &mut AstExpr) {
+        match expr {
+            AstExpr::StringLiteral(s) => rebuild_string_literal(s),
+            AstExpr::FString(f) if f.value.is_implicit_concatenated() => merge_f_string_parts(f),
+            _ => {}
+        }
+        walk_expr(self, expr);
     }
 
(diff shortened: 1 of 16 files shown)


crates/monty/test_cases/import__future_unknown_feature.py
davidhewitt, line 1: Is there a way we could make this work without the future import? I guess the problem is the test cases, so we'd have to go straight to the full evaluated annotations to match CPython?

crates/monty/test_cases/refcount__with_class.py
```diff
@@ -25,4 +25,7 @@ def __exit__(self, typ, val, tb):
 # binding (refcount 2); `cm` and `bound` alias the one instance (refcount 2);
 # `CM` is held by the global plus the instance's class slot (refcount 2).
 survived = cm.last_exc
-# ref-counts={'CM': 2, 'cm': 2, 'bound': 2, 'survived': 2}
+# The class's synthesized `__annotations__` dict (empty here) is a heap object
+# owned by the class namespace, so bind it too — see `refcount__class.py`.
+ann = CM.__annotations__
+# ref-counts={'CM': 2, 'cm': 2, 'bound': 2, 'survived': 2, 'ann': 2}
```
davidhewitt, line 31: I don't think we need this in this file, think this test was purely about the `with` statement refcounting.


```suggestion
```

crates/monty/tests/parse_errors.rs
```diff
@@ -126,6 +126,85 @@ fn unknown_imports_compile_successfully_error_deferred_to_runtime() {
...
+    // so rejecting them would fail class bodies that are perfectly valid.
+    for body in [
+        "__annotations__ = {'a': 'b'}",
+        "def __annotations__(self):\n        return 1",
+    ] {
+        let code = format!("class C:\n    {body}\n");
+        let result = MontyRun::new(code, "test.py", vec![], CompileOptions::default());
+        assert!(result.is_ok(), "`{body}` should compile");
+    }
+}
```
davidhewitt, line 155: The comment and name doesn't seem to match the test content?

crates/monty/src/stringize.rs
davidhewitt, line 1: If we went straight for PEP 649, would we need this code?

davidhewitt (commented):
A powerful step towards dataclasses!

My main question is whether we could implement a (possibly limited form of) PEP 649's lazy evaluation now, and skip the need for all the `__future__` and stringizing machinery?

crates/monty/tests/parse_errors.rs
rewitt94: Thank you, updated

crates/monty/test_cases/refcount__with_class.py
rewitt94: Thanks, removed!

crates/monty/test_cases/import__future_unknown_feature.py
rewitt94: The futures import is meaningless when we only have one implementation -

In CPython - 
with `from __future__ import annotations` -> PEP 563 (stringized)
with no`from __future__ import annotations ` -> PEP 649 (not deferred evaluation using descriptors)

In monty (with this PR) regardless -> stringized PEP 563 

It's valid python Syntax so we should probably parse it instead of throw a module not found?

crates/monty/src/stringize.rs
rewitt94: In CPython - 
with `from __future__ import annotations` -> stringized PEP 563 

So it is the backwards compatible case, but maybe we would want to support stringized anyway

rewitt94:
In short, the path to PEP 649 was very complicated and I wanted to deliver dataclasses sooner, I thought best to go with the stringized approach for now (especially if it's supported by CPython regardless).

i.e. 
```
Monty code changes to implement PEP 649

- Compile annotations into a synthetic per-class __annotate__(format) function, not a static dict.
- Reuse the existing class-body synthetic-function machinery in parse_class_def for it.
- Capture enclosing scope in that function so annotations can reference locals.
- Make __annotations__ a lazy caching descriptor on the class rather than a member.
- Add types.GenericAlias — subscripting a type currently raises TypeError.
- Add types.UnionType — int | None currently raises TypeError.
- Add a types module; it doesn't exist today.
- Add annotationlib: Format, get_annotations, ForwardRef, call_annotate_function.
- Implement FORWARDREF recovery so undefined or unsupported annotations degrade instead of raising.
- Keep stringize.rs as the Format.STRING backend.
- Make from __future__ import annotations switch modules to stringizing, no longer a no-op.
```

Here are also a list of user facing differences for PEP 563 vs PEP 649
```
User-facing differences, 563 vs 649

- __annotations__ values are strings under 563, real objects under 649.
- __annotate__ is None under 563, a function under 649.
- Undefined names are harmless under 563; 649 raises NameError on access.
- 649 resolves locals and closure variables; 563 cannot, giving NameError.
- 649 raises annotation errors on first access; 563 never evaluates, so never raises.
- dataclasses.fields(D)[i].type is 'int' under 563, <class 'int'> under 649.
- Generated __init__.__annotations__ holds strings under 563.
- ClassVar/InitVar detection is textual under 563; locally-aliased ClassVar becomes a real field.
- get_type_hints() needs the right namespace under 563; works directly under 649.
- annotationlib FORWARDREF/STRING formats return plain strings under 563.
```

davidhewitt (approved):
Having spoken about this offline, this is what I've reached:

- The `from __future__ import X` is reasonable to support because we might want to support a future hypothetical CPython feature anyway one day
- So given that the rest are nops, it makes sense for `from __future__ import annotations` to give PEP 563 results

It is a little unfortunate that this makes the default behaviour even thought it diverges from CPython's, however I think in the long term we can build on this to get the proper semantics (and it unblocks dataclasses in the interim).

PR #594, opened 2026-07-21
make `ResourceTracker` no longer generic

davidhewitt:
Disclaimer: fully vibe coded, I want to get pipeline feedback, will clean up.

We expect some perfomance "regression" from benchmarks because we're adding resource tracking to the measurement. This is arguably better as it matches real-world usage.

---
## Summary by cubic
Removed the generic `ResourceTracker` from the VM and standardized on a single concrete tracker. `ResourceLimits` are now passed directly, and omitted limits use Monty’s finite defaults across Rust, JS, Python, and the wire protocol.

- **Refactors**
  - `VM<'h, T>` → `VM<'h>`; all `impl ResourceTracker` generics removed from APIs, builtins, and helpers.
  - Unified tracker usage: drop `LimitedTracker` and `NoLimitTracker`; pass `ResourceLimits` directly (e.g., `MontyRepl::new(..., ResourceLimits::new(), ...)`).
  - Default policy change: omitted limits retain finite defaults (memory, allocations, duration, recursion). Updated docs/comments in `monty`, `monty-js`, `monty-python`, and `monty-proto`.
  - API tweaks: `MontyRun::run_no_limits` → `run_default`; examples/tests/fuzz targets updated accordingly.
  - Proto: `ResourceLimits` fields are optional overrides; conversions saturate safely; worker enforces host-selected limits on load/restore.

- **Migration**
  - Rust: replace `NoLimitTracker`/`LimitedTracker` with `ResourceLimits::default()` or a configured `ResourceLimits`; update call sites to use `&mut VM<'_>` without a tracker generic.
  - JS/Python: omitted/None limit fields no longer mean “unlimited”; set explicit values if you relied on unbounded behavior.
  - Proto clients: if you previously omitted fields to disable limits, send explicit large values instead; host-side limits override serialized limits on session load.

<sup>Written for commit b644dd1b963e7452859ab51937811c579768f444. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/594?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/src/resource.rs
```diff
@@ -220,328 +221,139 @@ impl From<ResourceError> for RunError {
...
-///
-/// Tracks allocation count, memory usage, and execution time, returning
-/// errors when limits are exceeded. Also schedules garbage collection
-/// at configurable intervals.
-///
-/// Uses `Cell` for interior mutability to allow many methods which take
-/// `&self` (enabling `&self` on critical methods such as `Heap::allocate`).
+/// Tracks and enforces resources for one execution session.
 ///
-/// `max_duration` limits *cumulative execution time*: the clock runs only
```
samuelcolvin, line 489: A lot of docstrings got deleted. Is that intentional?

crates/monty/src/resource.rs
```diff
@@ -144,14 +137,20 @@ fn estimate_bits_to_bytes(bits: u64) -> usize {
 pub enum ResourceError {
     /// Maximum number of allocations exceeded.
     Allocation { limit: usize, count: usize },
+    /// The attempted allocation count is not representable.
+    AllocationUnrepresentable { limit: usize },
     /// Maximum execution time exceeded.
     Time { limit: Duration, elapsed: Duration },
+    /// The attempted cumulative execution duration is not representable.
+    TimeUnrepresentable { limit: Duration },
```
samuelcolvin, line 145: I don't entirely get the idea of this?

samuelcolvin (commented):
this is going to take a lot of changes since we've introduced `monty-types` since this was started. Maybe we should start again?

samuelcolvin:
replaced by #613.

PR #598, opened 2026-07-21
unify `ExtFunction` types

davidhewitt:
This PR makes object identity simpler / cheaper by removing `Value::ExtFunction`. Instead, all external function calls are heap allocated (they cross the host boundary anyway so the cost is negligible).

To maintain ID stability I introduced weak cache of ext function objects.

---
## Summary by cubic
Unifies external function handling by removing `Value::ExtFunction` and making all host callables heap-backed objects cached by name. Identity now uses object identity while live; fixes GC cache leaks and rebuilds the cache on snapshot load; snapshot format bumped to 6.

- **Refactors**
  - Added `types::ExtFunction` (Arc-backed) and a heap weak cache keyed by name; `heap.get_ext_function` reuses live objects.
  - Compiler/VM allocate external functions for undefined globals in call context; calls yield `External` with a cloned name.
  - Simplified identity: `Value::id()`/`is()` no longer take `VM`; external functions compare/hash by object identity while live; `id()` encodes via `LongInt::value_from_u128`.
  - Updated object bridge and REPL to always use heap-backed external functions; docs clarify live name-based identity.

- **Bug Fixes**
  - Removed stale cache entries before slot reuse and during GC; cache is rebuilt on snapshot load to preserve identity across dumps.

<sup>Written for commit 26dcb0947b21c752c369e5a75432b2f47a8ace00. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/598?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/tests/external_function_identity.rs
```diff
@@ -244,3 +205,97 @@ fn repl_cross_representation_extfunction_identity() {
...
+            PrintWriter::Stdout,
+        )
+        .unwrap();
+    let (_repl, result) = progress.into_complete().expect("feed 3 should complete");
+
+    assert_eq!(
+        result,
+        MontyObject::Tuple(vec![MontyObject::Bool(false), MontyObject::Bool(false)]),
+    );
+}
```
rewitt94, line 301: Here is a failing test case to look at please -

```
/// The cycle collector must evict the weak-cache entry, like `dec_ref` does.
///
/// `for_each_child_id` enumerates every `Value::Ref` child regardless of whether
/// the child is GC-tracked, so an external function reachable only from a garbage
/// cycle is freed by `collect_white`. If that path leaves the cache pointing at
/// the freed `HeapId`, a later lookup of the same name resurrects whatever object
/// has since been allocated into the recycled slot.
#[cfg(feature = "test-hooks")]
#[test]
fn gc_freed_ext_function_does_not_retain_freed_id() {
    let code = "\
import gc
d = {}
d['f'] = a
d['self'] = d
d = None
a = None
gc.collect()
occupied = [1, 2, 3]
y = missing
(y is occupied, repr(y))
";
    let runner = MontyRun::new(code.to_owned(), "test.py", vec!["a".to_owned()], CompileOptions::default()).unwrap();
    let progress = runner
        .start(
            vec![MontyObject::Function {
                name: "ext_fn".to_owned(),
                docstring: None,
            }],
            NoLimitTracker,
            PrintWriter::Stdout,
        )
        .unwrap();

    let lookup = progress.into_name_lookup().expect("expected NameLookup for 'missing'");
    assert_eq!(lookup.name, "missing");
    let progress = lookup
        .resume(
            NameLookupResult::Value(MontyObject::Function {
                name: "ext_fn".to_owned(),
                docstring: None,
            }),
            PrintWriter::Stdout,
        )
        .unwrap();
    let result = progress.into_complete().expect("run should complete");

    assert_eq!(
        result,
        MontyObject::Tuple(vec![
            MontyObject::Bool(false),
            MontyObject::String("<function 'ext_fn' external>".to_owned()),
        ]),
    );
}
```

limitations/builtins.md
```diff
@@ -74,17 +74,12 @@ mechanism beyond dataclass field inheritance.
   Monty returns `0` immediately.
 - **`print`** — writes via the host print callback. `file=`, `flush=` are
   not honoured; `sep=` and `end=` are.
-- **`id(f)` / `hash(f)` / `f is g` / `f == g` for host-supplied callables**
-  — host functions passed in as inputs (`MontyObject::Function`) lose their
-  host object identity at the sandbox boundary. Inside Monty they are
-  compared by `__name__` alone: two distinct host callables with the same
-  `__name__` satisfy `a is b`, `a == b`, `id(a) == id(b)`, and
-  `hash(a) == hash(b)`. In CPython those would be four separate objects
```
rewitt94, line 82: Is it useful to keep documentation of what CPython does also for the comparison?

rewitt94 (commented):
Cool to learn about external function calls! Good idea to unify these 👍

limitations/builtins.md
davidhewitt: I think given CPython doesn't have "external functions" we probably don't care about any "compatible" semantics for them.

crates/monty/src/heap.rs
```diff
@@ -1052,6 +1096,17 @@ impl<T: ResourceTracker> Heap<T> {
...
                         reader.heap.purple_count -= 1;
                     }
+                    // Remove weak-cache entries before the slot becomes available for reuse.
+                    let ext_function_name = match ptr.data(reader) {
+                        HeapData::ExtFunction(function) => Some(function.cache_key()),
+                        _ => None,
+                    };
+                    // Clear the cache (only if it points to this exact function, it's possible for
+                    // snapshot deserialization to create duplicate functions with the same name)
+                    if let Some(name) = ext_function_name {
```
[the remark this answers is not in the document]
davidhewitt: I think it's an edge case from a bad snapshot, garbage-in-garbage-out applies here.

crates/monty/src/heap.rs
```diff
@@ -1224,6 +1279,10 @@ impl<T: ResourceTracker> Heap<T> {
                 heap_entry.readers.get(),
             );
             let mut value = entry.free();
+            // Clear weak entries before freeing their slots, just as `dec_ref`.
+            if let HeapData::ExtFunction(function) = value.data.0.get_mut() {
+                Self::remove_ext_function_cache_entry(&mut self.ext_function_cache, &function.cache_key(), id);
```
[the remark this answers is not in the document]
davidhewitt: As per https://github.com/pydantic/monty/pull/598#discussion_r3655640073 I am not concerned.

PR #608, opened 2026-07-22
Add `collections`: deque, namedtuple, defaultdict, Counter

rewitt94:
Makes `collections` importable and implements four of its types.

- **`deque`** — a new heap type backed by `VecDeque` (`types/deque.rs`): all 14 methods, the read-only `maxlen` with left/right eviction, indexing/setitem, `==`/ordering, `+ * += *=`, repr, and CPython's mutation-during-iteration check.
- **`namedtuple(...)`** — a real class object, so `type(p) is Point` and `isinstance` work. Supports `rename`, `defaults`, `_fields`, `_field_defaults`, `_make`, `_replace`, `_asdict`.
- **`defaultdict`** and **`Counter`** — a `dict` carrying a `DictKind` discriminant rather than separate heap types, so they reuse dict's storage, iteration and refcount paths unchanged. Counter has `most_common`, `elements`, `total`, `update`/`subtract` and the `+ - & |` operators.

All four are exposed as the type objects themselves, not factory functions.

### Worth a closer look

**`heap.rs`** gains two `HeapData` variants (`Deque`, `NamedTupleClass`) and their GC trace arms. The namedtuple arm reports its class edge *before* the `contains_refs` early-out — that flag is computed from the instance's items alone, so an all-primitive `Point(1, 2)` would otherwise drop the class reference. No `unsafe` was touched.

**The mutation check is counter-based, not length-based.** `rotate()`, and a paired `append()`/`popleft()`, leave the length unchanged while invalidating iteration; CPython catches both and so does this.

**Type names.** Monty stores one name per type; CPython picks qualified or bare *per surface*, depending on whether the type is written in C or in Python. `deque` and `defaultdict` are C types, so they are named `collections.<name>` and match CPython's `repr` and every type-naming error message, diverging only on `__name__`. `Counter` is a Python-level class, so the bare name is the one that matches. This asymmetry is deliberate and documented.

**Typeshed.** The vendored `collections` stub is replaced by a narrowed custom one covering only the four implemented names, so `from collections import OrderedDict` is a type error rather than something that type-checks and then fails at runtime. `crates/monty/tests/collections.rs` pins the stub and the runtime together.

### Not implemented

`OrderedDict`, `ChainMap`, `UserDict`, `UserList`, `UserString`, and the `collections.abc` submodule.

Divergences from CPython are documented in `limitations/collections.md` and `limitations/namedtuple.md`; `limitations/builtins.md`, `language.md` and `modules.md` are updated where this work corrected or invalidated existing entries.

---
## Summary by cubic
Adds the `collections` module with `deque`, `namedtuple`, `defaultdict`, and `Counter`, matching CPython 3.14. Also routes unary/binary/in‑place ops through per‑type hooks and preserves the qualified `collections.deque` type across host/wire.

- **New Features**
  - `collections` exports `deque`, `namedtuple`, `defaultdict`, `Counter`; `namedtuple(...)` returns a class and calling it constructs an instance; the qualified name `collections.deque` round‑trips through the host.
  - `deque`: full API including `maxlen`, indexing/assignment, `+`/`*` and in‑place ops, reversed iteration, mutation‑during‑iteration checks, and a dedicated iterator.
  - `defaultdict`/`Counter`: reuse `dict` via a `DictKind` tag; `Counter` adds `most_common`, eager `elements` (returns `list`), `total`, `update`/`subtract`, unary `+`/`-`, and `+`/`-`/`&`/`|` plus multiset comparisons (`<=`/`>=`).
  - VM/typing/tooling: unary/binary/in‑place operator dispatch moved to per‑type hooks (incl. Counter’s multiset comparisons); `+timedelta` is an identity; narrowed `collections` typeshed with nested custom stubs; REPL function lists exclude classes/`NamedTupleClass`/bound methods; `str.isidentifier()` now follows full Unicode rules.

- **Bug Fixes**
  - Deque: ordering falls back to equality for unorderable‑equal elements; `extend`/`extendleft` append while consuming (self‑extend snapshots); repetition bounds memory, polls time, preserves `maxlen`, and releases clones on timeout; `+`/`+=` use CPython’s “can only concatenate deque …” wording; `reversed` recognizes `deque`.
  - Dict‑backed types: moved missing‑key subscript handling to per‑type hooks — `defaultdict` inserts via its factory and releases its read handle before calling it; `Counter` reads `0` without inserting.
  - Counter: bounds `elements()` by memory; `most_common(n)` accepts `bool`/big‑int; fixes multiset comparisons (incl. `NaN`) and in‑place op acceptance; uses drop guards; `counter_bump` now guards the key and borrows the delta.
  - Namedtuple: correct error quoting/escaping; class teardown releases heap‑ref defaults; `type()`/`isinstance()` match the factory class; `sys.version_info` retains its structseq surface only.

<sup>Written for commit 9c3b9bb7e86c7d898988a0e71a2af053b70eb188. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/608?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/test_cases/collections__counter.py
```diff
@@ -0,0 +1,339 @@
...
+    assert False, 'expected non-iterable to raise'
+except TypeError as e:
+    assert str(e) == "'int' object is not iterable", 'non-iterable message'
+
+# An unhashable element raises TypeError, and the counter (plus the elements
+# after the bad one) must not leak — regression for counter_bump / counter_update.
+# Only the type is asserted: Monty's unhashable-key message is a pre-existing
+# dict divergence from CPython's, so the exact text can't be dual-run.
+try:
+    Counter([1, [2], 3])
```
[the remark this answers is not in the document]
rewitt94: Assertion added in `crates/monty/test_cases/refcount__collections.py`

```
unhashable_key = Box()
try:
    Counter([unhashable_key, [1]])   # [1] is unhashable → raises mid-build
    assert False, 'expected an unhashable element to raise'
except TypeError:
    pass
```

crates/monty-typeshed/custom/collections/__init__.pyi
```diff
@@ -0,0 +1,190 @@
...
+    def __mul__(self, value: int, /) -> Self: ...
+    def __imul__(self, value: int, /) -> Self: ...
+    def __lt__(self, value: deque[_T], /) -> bool: ...
+    def __le__(self, value: deque[_T], /) -> bool: ...
+    def __gt__(self, value: deque[_T], /) -> bool: ...
+    def __ge__(self, value: deque[_T], /) -> bool: ...
+    def __eq__(self, value: object, /) -> bool: ...
+    def __class_getitem__(cls, item: Any, /) -> GenericAlias: ...
+
+class Counter(dict[_T, int], Generic[_T]):
```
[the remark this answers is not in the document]
rewitt94: This mirrors upstream typeshed's Counter typing and Monty behaves identically to CPython at runtime (both accept non-int counts). I don't think we should diverge on the typestub here.

crates/monty/src/types/deque.rs
```diff
@@ -0,0 +1,811 @@
...
+            deque.appendleft(vm, item)?;
+            Ok(Value::None)
+        }
+        StaticStrings::Pop => {
+            args.check_zero_args("deque.pop", vm.heap)?;
+            let this = deque.get_mut(vm.heap);
+            // An empty pop raises without mutating, so it must not bump the state.
+            let item = this
+                .items
+                .pop_back()
```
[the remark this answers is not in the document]
rewitt94: This is deliberate, repo-wide behaviour `list`/`dict`/`set` don't debit the tracker on explicit removal either. I thought best not to break convention here.

crates/monty/src/heap_data.rs
```diff
@@ -1163,3 +1226,18 @@ impl<'h> PyTrait<'h> for HeapReadOutput<'h> {
...
+/// degrades to a plain `tuple` under `+` (as in CPython), and the operands must be
+/// cloned before allocating because `allocate_tuple` cannot run while two
+/// `HeapRead` handles are live.
+fn concat_into_tuple(
+    a: Vec<Value>,
+    b: Vec<Value>,
+    vm: &mut VM<'_, impl ResourceTracker>,
+) -> Result<Option<Value>, ResourceError> {
+    let items: TupleVec = a.into_iter().chain(b).collect();
+    Ok(Some(allocate_tuple(items, vm.heap)?))
```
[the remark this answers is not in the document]
rewitt94: The allocation path here can only raise Memory/Time which both terminal so currently not an issue. I think in similar instances in the repo it is left like-so.

crates/monty/test_cases/collections__namedtuple.py
```diff
@@ -0,0 +1,380 @@
...
+
+# === Heap-reference default values (exercises the class object's refcounting) ===
+ListDefault = namedtuple('ListDefault', ['a', 'b'], defaults=[[9]])
+assert ListDefault(0) == (0, [9]), 'heap-ref default applied to an omitted field'
+assert ListDefault(0, 1) == (0, 1), 'provided value overrides the heap-ref default'
+# Freeing a class whose default holds a heap reference must release it (regression
+# for a missing NamedTupleClass arm in the GC walkers).
+_tmp = namedtuple('Tmp', ['a'], defaults=[[1, 2]])
+assert _tmp() == ([1, 2],), 'temporary class heap-ref default applied'
+_tmp = 0
```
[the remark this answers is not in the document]
rewitt94: Now covered by the `NtDefaults` case in `refcount__collections.py`

crates/monty-typeshed/custom/collections/__init__.pyi
```diff
@@ -0,0 +1,190 @@
...
+_KT = TypeVar('_KT')
+_VT = TypeVar('_VT')
+
+# namedtuple is special-cased in the type checker; the initializer is ignored.
+def namedtuple(
+    typename: str,
+    field_names: str | Iterable[str],
+    *,
+    rename: bool = False,
+    module: str | None = None,
```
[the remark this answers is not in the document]
rewitt94: Again runtime matches CPython and typestub matches upstream typestub

crates/monty-typeshed/custom/collections/__init__.pyi
```diff
@@ -0,0 +1,190 @@
...
+_S = TypeVar('_S')
+_T = TypeVar('_T')
+_T1 = TypeVar('_T1')
+_T2 = TypeVar('_T2')
+_KT = TypeVar('_KT')
+_VT = TypeVar('_VT')
+
+# namedtuple is special-cased in the type checker; the initializer is ignored.
+def namedtuple(
+    typename: str,
```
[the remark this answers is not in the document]
rewitt94: Again runtime matches CPython and typestub matches upstream typestub

crates/monty/src/modules/collections/counter.rs
```diff
@@ -0,0 +1,873 @@
...
+
+/// `Counter.update`/`subtract` method body: binds `iterable=None, /, **kwargs`
+/// and folds the counts in. Returns `None`.
+pub(crate) fn counter_update_method(
+    dict_id: HeapId,
+    args: ArgValues,
+    subtract: bool,
+    vm: &mut VM<'_, impl ResourceTracker>,
+) -> RunResult<Value> {
+    let (mut pos, kwargs) = args.into_parts();
```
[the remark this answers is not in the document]
rewitt94: Unfortunately, there is a mismatch in the arguments index because CPython counts the implicit self and FromArgs has no way to represent it so that approach does not work here.

crates/monty/src/value.rs
```diff
@@ -2570,6 +2667,73 @@ fn py_float_mod(a: f64, b: f64) -> f64 {
...
+/// window of `L = min(len*count, maxlen)` items starts at `(len - L % len) % len`
+/// within the pattern and wraps. An unbounded deque has no such shortcut and
+/// materializes the full product (which may exhaust resource limits, matching
+/// CPython's `MemoryError` on an impossibly large repeat).
+fn repeat_deque(deque: &Deque, count: usize, vm: &VM<'_, impl ResourceTracker>) -> RunResult<Value> {
+    let len = deque.len();
+    let result = if let Some(max) = deque.maxlen() {
+        // Bounded: keep only the last `min(len*count, maxlen)` items.
+        let kept = len.saturating_mul(count).min(max);
+        let mut result = Vec::with_capacity(kept);
```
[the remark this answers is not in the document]
rewitt94: Fixed in `d2d02c24`

crates/monty-python/tests/test_types.py
```diff
@@ -156,6 +159,8 @@ def test_type_object_output(monty_run: RunMonty):
         datetime.timezone,
         re.Pattern,
         re.Match,
+        # deque lives in `collections`, not `builtins` — resolves via its own arm.
```
davidhewitt: I have no idea what this comment means which makes me think I want it gone 😂 

```suggestion
```

crates/monty/src/builtins/reversed.rs
```diff
@@ -61,6 +61,9 @@ fn is_reversible(value: &Value, vm: &VM<'_>) -> bool {
             HeapData::List(_)
                 | HeapData::Tuple(_)
                 | HeapData::NamedTuple(_)
+                // A deque defines `__reversed__` in CPython; sets deliberately
+                // stay out, having no order to reverse.
```
davidhewitt:
```suggestion
```

crates/monty/src/bytecode/vm/binary.rs
davidhewitt, line 1: I don't think this file should be edited; this logic should be routed onto the correct types via `PyTrait`.

crates/monty/src/bytecode/vm/compare.rs
davidhewitt, line 1: Again I think this should go via `PyTrait` rather than editing this file.

crates/monty/src/bytecode/vm/mod.rs
davidhewitt: Looks like `py_neg` should be a method on `PyTrait` and we should extract the type-level logic out of this file.

crates/monty/src/modules/collections/counter.rs
```diff
@@ -0,0 +1,1063 @@
...
+            // `_count_elements`: `self[elem] = self_get(elem, 0) + 1`, so the
+            // existing count is the left operand here.
+            let items = collect_owned_iterable::<Vec<Value>>(source, vm)?;
+            let mut items = items.into_iter();
+            while let Some(item) = items.next() {
+                if let Err(e) = counter_bump(dict_id, item, Value::Int(1), subtract, false, vm) {
+                    // A bump can fail on an unhashable item — drop the rest.
+                    for rest in items {
+                        rest.drop_with(vm);
+                    }
```
davidhewitt: looks like some `defer_drop_mut!` probably missing.

crates/monty/src/modules/collections/counter.rs
```diff
@@ -0,0 +1,1063 @@
...
+        unreachable!("counter_bump on a non-dict heap entry");
+    };
+    // `dict_get` hashes the key and can fail for an unhashable key (e.g.
+    // `Counter([[1]])`); own that error path and drop the key and delta.
+    let base = match dict.dict_get(&key, vm) {
+        Ok(Some(existing)) => existing,
+        Ok(None) => Value::Int(0),
+        Err(e) => {
+            key.drop_with(vm);
+            delta.drop_with(vm);
```
davidhewitt: `DropGuard` on key to potentially transfer ownership at the end, looks like `delta` should be `&value`.

crates/monty/src/modules/collections/counter.rs
```diff
@@ -0,0 +1,1063 @@
...
+/// `Counter.update`/`subtract` method body: binds `iterable=None, /, **kwargs`
+/// and folds the counts in. Returns `None`.
+///
+/// The binding is hand-rolled rather than `#[derive(FromArgs)]`: these are
+/// pure-Python `def` methods, so CPython's over-arity error counts the implicit
+/// `self` (`Counter.update() takes from 1 to 2 positional arguments but 3 were
+/// given`), which the derive's `def` family — positional-only, no receiver —
+/// cannot express. The `+ 1`s below re-introduce that `self`.
+pub(crate) fn counter_update_method(
+    dict_id: HeapId,
```
davidhewitt, line 299: It seems to be that `dict_id` is frequently `.read()` and not much else, might be better to pass `HeapReader<'_, Dict>` around in this file? Not sure.

crates/monty/src/modules/collections/counter.rs
```diff
@@ -0,0 +1,1063 @@
...
+        count.drop_with(vm);
+        match next {
+            Ok(next) => sum = next,
+            Err(e) => {
+                for rest in counts {
+                    rest.drop_with(vm);
+                }
+                return Err(e);
+            }
+        }
```
davidhewitt: this is claude making match spaghetti because it can't do drop guards right, almost certainly `DropGuard` on `sum` and `defer_drop_mut!` on `counts`.

crates/monty/src/types/py_trait.rs
davidhewitt: Nope we should change the return type to `RunResult<bool>` so it can be here.

crates/monty/src/object_bridge.rs
```diff
@@ -606,6 +629,9 @@ impl MontyTypeExt for MontyType {
             Type::Tuple => Self::Tuple,
             Type::NamedTuple => Self::NamedTuple,
             Type::Dict => Self::Dict,
+            // No host-side defaultdict/Counter type; both degrade to `dict`,
+            // consistent with their values crossing as `MontyObject::Dict`.
+            Type::DefaultDict | Type::Counter => Self::Dict,
```
davidhewitt, line 644: Is this correct? Seems like Python could move `Counter` across faithfully at the very least.

crates/monty/src/value.rs
```diff
@@ -1162,7 +1187,28 @@ impl<'h> PyTrait<'h> for Value {
...
+                            HeapReadOutput::Dict(dict) => dict.dict_get(key, vm)?,
+                            _ => unreachable!("defaultdict/Counter is a dict"),
+                        };
+                        return match found {
+                            Some(value) => Ok(value),
+                            None if is_defaultdict => defaultdict_missing(id, key, vm),
+                            None => Ok(Self::Int(0)),
+                        };
+                    }
+                }
```
davidhewitt: Seems like this logic should go in `heap_data.rs`?

crates/monty/src/bytecode/vm/binary.rs
rewitt94: Agreed, nothing type specific in here, although I have done some structural refactoring now.

crates/monty/src/bytecode/vm/compare.rs
rewitt94: Good spot, thanks

crates/monty/src/bytecode/vm/mod.rs
rewitt94: Did both `py_neg` and `py_pos`

crates/monty/src/value.rs
rewitt94: Moved via a `heap_subscript` function

crates/monty/src/modules/collections/counter.rs
rewitt94: True... I think that would add other complications though

crates/monty/src/modules/collections/counter.rs
```diff
@@ -0,0 +1,1050 @@
...
+/// so an unorderable count reports `>` and a `NaN` count is not positive.
+fn count_is_positive(value: &Value, vm: &mut VM<'_>) -> RunResult<bool> {
+    count_holds(value, &Value::Int(0), CountCmp::Gt, vm)
+}
+
+/// Converts a count to a repetition length for `elements()`.
+///
+/// CPython feeds the count to `itertools.repeat`, which requires a whole
+/// number, so a float count raises rather than being truncated. Negative counts
+/// yield `0` (the element is skipped).
```
davidhewitt, line 125: Should there be a `TODO` here to use the itertools machinery here once that merges?

crates/monty/src/modules/collections/counter.rs
```diff
@@ -0,0 +1,1050 @@
...
+/// The right operand `rhs` may be **any mapping** (`c += {'a': 2}`), not only a
+/// Counter. `+=`/`-=`/`|=` iterate `rhs.items()`, so a non-mapping raises
+/// `AttributeError`; `&=` subscripts `rhs[elem]`, so a plain dict raises
+/// `KeyError` for a key absent from it (a Counter reads that as `0`).
+///
+/// Note the asymmetry CPython has between the directions: `+=` / `-=` walk
+/// *other*'s keys (so new keys can appear), whereas `&=` walks *self*'s keys
+/// (so no key is ever added), and `|=` walks *other*'s keys keeping the larger.
+pub(crate) fn counter_inplace_op(l_id: HeapId, rhs: &Value, op: CounterOp, vm: &mut VM<'_>) -> RunResult<()> {
+    match op {
```
davidhewitt, line 784: I wonder if here we should be doing `vm.heap.read(l_id)` and asserting we have a counter immediately, and passing `HeapRead<'h, Dict>` down to the inner functions to avoid re-read?

crates/monty/src/modules/collections/counter.rs
```diff
@@ -0,0 +1,1050 @@
...
+
+/// Drops every remaining `(Value, Value)` pair from an iterator, debiting the
+/// heap refcount each holds. Used to release the un-consumed tail after an early
+/// exit, where a plain `Drop` of the iterator would leak (it never sees the VM).
+fn drain_pairs(pairs: impl Iterator<Item = (Value, Value)>, vm: &mut VM<'_>) {
+    for (a, b) in pairs {
+        a.drop_with(vm);
+        b.drop_with(vm);
+    }
+}
```
davidhewitt, line 717: There is a `DropWithContext` implementation for `vec::IntoIter<T>` as well as for 2-tuples so I'm reasonably sure that anywhere this is present we can use `defer_drop_mut!` on the iterator to hopefully simplify code paths massively.

crates/monty/src/types/deque.rs
davidhewitt, line 1: Surely this should go in `collections/deque.rs`?

crates/monty/src/heap_data.rs
```diff
@@ -203,6 +217,9 @@ impl HeapData {
             | Self::Instance(_)
             | Self::BoundMethod(_)
             | Self::ListIterator(_)
+            // A deque iterator owns a reference to its deque (a container), so it
+            // can close a cycle exactly like `ListIterator`.
```
davidhewitt:
```suggestion
```

crates/monty/src/heap_data.rs
```diff
@@ -187,8 +196,13 @@ impl HeapData {
         match self {
             Self::Itertools(iter) => iter.is_gc_tracked(),
             Self::List(_)
+            | Self::Deque(_)
             | Self::Tuple(_)
             | Self::NamedTuple(_)
+            // A namedtuple class owns its `defaults`, which may be heap refs
+            // (e.g. `namedtuple('C', ['a'], defaults=[[]])`), so it can sit on
+            // a cycle just like `Class`.
```
davidhewitt:
```suggestion
```

crates/monty/src/heap_data.rs
```diff
@@ -259,7 +276,10 @@ impl HeapData {
             Self::Str(_) => Type::Str,
             Self::Bytes(_) => Type::Bytes,
             Self::List(_) => Type::List,
+            Self::Deque(_) => Type::Deque,
             Self::Tuple(_) | Self::NamedTuple(_) => Type::Tuple,
+            // A namedtuple class object's type is `type`, like any class.
```
davidhewitt:
```suggestion
```

crates/monty/src/heap_data.rs
```diff
@@ -984,7 +1072,9 @@ impl<'h> PyTrait<'h> for HeapReadOutput<'h> {
             Self::DateTime(value) => value.py_iter(self_id, vm),
             Self::TimeDelta(value) => value.py_iter(self_id, vm),
             Self::TimeZone(value) => value.py_iter(self_id, vm),
-            Self::Closure(_)
+            // A namedtuple *class* is a callable, not an iterable.
```
davidhewitt:
```suggestion
```

davidhewitt (approved):
I suggest we merge this now and then action a cleanup pass as a followup. See below comments.

I will apply the deletions and leave the rest for the followup.

PR #609, opened 2026-07-22
Dispatch user-defined `__iter__` / `__next__` / `__contains__`

rewitt94:
A class implementing the iterator protocol is now iterable everywhere — `for`,
`list()`/`sum()`/`sorted()`/`set()`, comprehensions, `in`, every unpacking form,
`join`, and `next()` — rather than raising `'C' object is not iterable`. The
type checker already accepted such a class, so this closes a `-t`-passes /
runtime-fails divergence rather than opening one.

`Instance` answers four `PyTrait` methods: `py_is_iterable`, `py_is_iterator`,
`py_iter` and `py_next`. The rest follows from #588 (every iteration callsite
routed through `py_iter`/`py_next`) and #589 (`py_is_iterable`). `MontyIter` is
untouched — once `py_iter` is overridden a user instance never reaches it.

Notable points:

- `py_iter` returns whatever `__iter__` returned, no wrapper interposed, so
  `iter(obj) is obj` holds for a self-iterator. The result must be an iterator,
  mirroring `PyObject_GetIter`'s `PyIter_Check` — that is what `py_is_iterator`
  is for.
- `py_next` gains `self_id: Option<HeapId>` because binding `self` needs the id
  and `HeapRead` does not carry one. `ValueRead::py_next` supplies it, so the
  ~30 builtin callers are untouched.
- `StopIteration` becomes exhaustion via `RunError::is_stop_iteration`, which
  matches only the catchable variant so a resource limit still propagates. The
  same helper fixes `iter(callable, sentinel)` — one bullet removed from
  `limitations/iter.md`.
- `str.join`/`bytes.join` rewrote any `into_py_iter` failure as `can only join
  an iterable`; now that a user `__iter__` can raise, they check
  `py_is_iterable` up front so `''.join(Boom())` surfaces the real error.

## `__contains__`

Dispatching `__iter__` would otherwise make a *wrong answer* reachable: `in`
never consulted `__contains__`, so a class defining both would silently return
the iteration answer where it previously raised. The last commit dispatches it
instead — `sq_contains` before `tp_iter`, with the result coerced by truthiness
as `PyObject_IsTrue` does.

`instance_call_dunder_sync` takes an optional argument to serve one-arg dunders,
and drops it itself when the class does not define the dunder, so no caller has
a path on which the clone can leak. The type checker needs nothing: `in` is an
operator, so it resolves against the class's own `__contains__` — verified with
a temporary probe in `good_types.py`, not assumed.

## Testing

`iter__user_class.py` covers every consumption site, both `__iter__` shapes and
each error path, byte-identical to CPython 3.14. `iter__reentrant.py` adds a
`__next__` that re-enters the object being advanced and one that mutates it
mid-iteration. `class__contains.py` covers precedence (the two answers are
deliberately opposite, so a fallback regression fails loudly), truthiness,
exception propagation, a non-callable `__contains__` and class-only lookup.

The two `refcount__*` cases prove nothing is retained; both were checked to
discriminate by breaking their expectations. Green under `memory-model-checks`
and `ref-count-return`. Suspension is a Rust-side test, since on CPython the
external is a real function and the call would succeed.

## Divergences

`limitations/classes.md`: `__iter__`/`__next__`/`__contains__` run
synchronously, so one calling an external/OS function raises
`NotImplementedError`; the `__getitem__`-only iteration fallback and
`__reversed__` remain undispatched. The `__contains__` entry is deleted rather
than reworded — the divergence is gone, not reshaped.
`limitations/exceptions.md`: a synchronously-invoked callable omits the calling
frame from its traceback — pre-existing, now also reachable here.

---
## Summary by cubic
Adds runtime dispatch for user-defined `__iter__`, `__next__`, and `__contains__` with CPython-accurate precedence, opt-outs, and error messages. Also upgrades refcount tests to detect real leaks via a transitive reachability walk.

- New Features
  - `in` calls `__contains__` first; `__contains__ = None` raises "'T' object is not a container". If absent, it iterates any iterable; non-containers raise "argument of type 'T' is not a container or iterable".
  - `__iter__ = None` opts out everywhere with "'T' object is not iterable". Unpacking and `*`-unpacking keep the plain wording for opted-out classes.
  - `iter(obj)` returns the user result unchanged. If it’s not an iterator, raise "iter() returned non-iterator of type 'T'" (via `py_is_iterator`).
  - `str.join`/`bytes.join` pre-check iterability so user `__iter__` errors surface. `__next__` is re-entrant-safe: `py_next(self_id, ...)` turns `StopIteration` into exhaustion; resource-limit errors propagate.

- Bug Fixes
  - `iter(callable, sentinel)` treats `StopIteration` from the callable as clean exhaustion.
  - `in` now falls back to iteration for any iterable (e.g. `callable_iterator`), not just a fixed set.
  - Unpacking errors distinguish opt-outs: opted-out classes keep "'T' object is not iterable" instead of site-specific wording.
  - Refcount strict check walks reachability from named globals; failures list unreachable heap objects by type/id, and tests no longer need extra bindings.

<sup>Written for commit 1380d9ca933dbaa1fb81f44b64495ee0d2c6c919. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/609?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/run.rs
+++ b/crates/monty/src/run.rs
@@ -449,15 +449,13 @@ impl Executor {
     /// Executes the code and returns both the result and reference count data with a custom tracker,
     /// used for testing only.
     ///
-    /// This is used for testing reference counting behavior with a custom tracker. Returns:
-    /// - The execution result (`Exit`)
-    /// - Reference count data as a tuple of:
-    ///   - A map from variable names to their reference counts (only for heap-allocated values)
-    ///   - The number of unique heap value IDs referenced by variables
-    ///   - The total number of live heap values
+    /// This is used for testing reference counting behavior with a custom tracker. Returns
+    /// the execution result plus, in [`RefCountOutput`], a map from variable names to their
+    /// reference counts (heap-allocated values only), any live-but-unreachable heap entries,
+    /// and the total live heap population.
     ///
-    /// For strict matching validation, compare unique_refs_count with heap_entry_count.
-    /// If they're equal, all heap values are accounted for by named variables.
+    /// For strict-matching validation, assert that `unreachable` is empty: every live heap
+    /// object should be reachable from a named variable, so anything left over is a leak.
     ///
     /// Only available when the `ref-count-return` feature is enabled.
     #[cfg(feature = "ref-count-return")]
@@ -466,8 +464,6 @@ impl Executor {
         inputs: Vec<MontyObject>,
         resource_tracker: ResourceTracker,
     ) -> Result<RefCountOutput, MontyException> {
-        use std::collections::HashSet;
-
         let mut heap = Heap::new(self.namespace_size(), resource_tracker);
         let globals = self.empty_globals();
 
@@ -492,7 +488,7 @@ impl Executor {
             // Read refcounts BEFORE converting the return value, because
             // `frame_exit_to_object` drops the return value (decrementing its refcount).
             let mut counts = ahash::AHashMap::new();
-            let mut unique_ids = HashSet::new();
+            let mut roots = Vec::new();
 
             for (namespace_id, name_id) in executor.globals.iter() {
                 let idx = namespace_id.index();
@@ -500,10 +496,17 @@ impl Executor {
                     && let Value::Ref(id) = &globals[idx]
                 {
                     counts.insert(executor.interns.get_str(name_id).to_owned(), vm.heap.get_refcount(*id));
-                    unique_ids.insert(*id);
+                    roots.push(*id);
                 }
             }
-            let unique_refs = unique_ids.len();
+            // Named globals are the only roots: locals are gone once the module frame
+            // exits, so anything still live must hang off a name to not be a leak.
+            let unreachable: Vec<String> = vm
+                .heap
+                .unreachable_entries(roots)
+                .into_iter()
+                .map(|(id, ty)| format!("{} (id {})", ty.name(vm.heap, &executor.interns), id.index()))
+                .collect();
             let heap_count = vm.heap.entry_count();
 
             // Convert return value while VM is still alive (needs access to interns).
@@ -521,7 +524,7 @@ impl Executor {
             Ok(RefCountOutput {
                 py_object,
                 counts,
-                unique_refs,
+                unreachable,
                 heap_count,
                 allocations_since_gc,
             })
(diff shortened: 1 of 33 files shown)


crates/monty/src/types/callable_iterator.rs
```diff
@@ -67,6 +67,10 @@ impl HeapItem for CallableIterator {
 }
 
 impl<'h> PyTrait<'h> for HeapRead<'h, CallableIterator> {
+    fn py_is_iterator(&self, _: &VM<'h>) -> bool {
+        true
+    }
+
     fn py_is_iterable(&self, _: &VM<'h>) -> bool {
         true
     }
```
davidhewitt, line 76: I think this works for now, however I wonder if long term we need proper "method resolution" for `py_next` and `py_iter` which could replace these with correct logic automatically. Not sure exactly what that looks like.

crates/monty/src/value.rs
```diff
@@ -1758,6 +1778,9 @@ impl Value {
     /// - Dict: key lookup
     /// - Set/FrozenSet: element lookup
     /// - Str: substring search
+    /// - Instance: the class's `__contains__`, else iteration
+    ///
+    /// Anything else iterable is consumed until the item is found.
     pub fn py_contains(&self, item: &Self, vm: &mut VM<'_>) -> RunResult<bool> {
```
davidhewitt, line 1502: It looks like we should move `py_contains` onto `PyTrait` for better location of type-specific logic. Would be great to note that for a follow-up please.

crates/monty/src/run.rs
```diff
@@ -492,18 +488,25 @@ impl Executor {
...
-                    unique_ids.insert(*id);
+                    roots.push(*id);
                 }
             }
-            let unique_refs = unique_ids.len();
+            // Named globals are the only roots: locals are gone once the module frame
+            // exits, so anything still live must hang off a name to not be a leak.
+            let unreachable: Vec<String> = vm
+                .heap
+                .unreachable_entries(roots)
```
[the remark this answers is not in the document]
rewitt94: I don't think this is related to this change, I will raise a different PR for this.

crates/monty/src/value.rs
rewitt94: I will raise a follow-up

davidhewitt (approved):
Plz fix merge conflicts, then let's ship!

davidhewitt (approved):

PR #620, opened 2026-07-24
Track iterator types for GC so their cycles can't evade collection

rewitt94:
## The bug

`ListIterator` and `CallableIterator` hold owned references to arbitrary user objects, but neither was listed in `HeapData::is_gc_tracked`. This means that they did not properly participate in cyclic GC checking.

Also changed the `is_gc_tracked` match to be exhaustive so that newly added HeapDate types in future are not missed.

Developed with TDD - failing test case now passes.


davidhewitt (approved):
Thanks

PR #626, opened 2026-07-27
Native `@dataclass` defined inside the sandbox

rewitt94:
Sandboxed Python can define its own dataclasses, executed entirely in-sandbox — distinct from host-supplied dataclasses, which dispatch back to the host.

**Design:** metadata, not codegen. The decorator records a `DataclassMeta` on the `Class`; dunders are synthesized off it. No `exec`, no generated bytecode. The decorated class stays an ordinary class, so methods, `isinstance`, `type(p) is Point` and `__class__` keep working. Logic is centralised in `modules/dataclasses.rs`, reached through three hooks: construction, equality dispatch, value dunders.

**Works:** `import dataclasses`, bare `@dataclass`, `is_dataclass`, field discovery from `__annotations__` (ordered, `ClassVar` excluded), synthesized `__init__` / `__repr__` / `__str__` / field-wise `__eq__`, unhashable by default. Arity and keyword `TypeError`s are byte-exact vs CPython.

**User `__eq__`/`__hash__` are dispatched** for every class, built on #609's `instance_call_dunder_sync`. So a dataclass body may define either, and CPython's precedence — a body-defined dunder beats the synthesized one — falls out of dispatch order. Defining `__eq__` makes the class unhashable, as CPython does implicitly; without that, instances would compare by value but hash by identity, quietly breaking dicts and sets.

**Rejected at decoration time**, rather than quietly building a wrong class: mutable default, non-default-after-default, `__post_init__`, `InitVar`, and the `@dataclass(...)` keyword form. The mutable-default check rests on real hashability, so a default that is an instance of a class setting `__hash__ = None` or defining `__eq__` counts too. The first two match CPython's messages exactly, and field validation runs first so any class CPython also rejects reports CPython's message.

**Not in this PR:** the keyword form (frozen/order/hash), `field()`/`default_factory`, `__post_init__`/`InitVar`, `fields`/`asdict`/`astuple`/`replace`. Field inheritance is permanently out — Monty has no class inheritance.

Two divergences documented rather than fixed: Monty has no `NotImplemented`, so a user `__eq__` cannot decline and the reflected operand is never tried; and `a == a` stays true even when `__eq__` disagrees, because the identity short-circuit sits in `Value::py_eq_impl`, which serves both `==` and container comparison where CPython shortcuts only the latter.

Compatible dump change: new `StaticStrings` are appended, so only the fingerprint moves, no `VERSION` bump.

---
## Summary by cubic
Adds a native `dataclasses` module so sandboxed Python can define `@dataclass` classes in-sandbox with captured defaults, introspectable `__dataclass_fields__`, and synthesized dunders matching CPython. Also routes user `__eq__`/`__hash__` for all classes, fixes `==` to respect a user `__eq__`, and avoids leaks if a memory limit aborts decoration.

- **New Features**
  - Importable `dataclasses` module with `dataclass` and `is_dataclass`.
  - Defaults are captured at decoration; fields read like attributes (instance dict, then class). Function defaults bind to the instance.
  - `__dataclass_fields__` mapping on the class holds heap-backed `dataclasses.Field` objects (introspectable). `Field` is a first-class Monty type.
  - Synthesized dunders: recursion-guarded, timeout-aware `__repr__`/`__str__`; field-wise `__eq__`; CPython-like hashing rules. Unsupported options (`@dataclass(...)`, `__post_init__`, `InitVar`) raise `NotImplementedError`.

- **Refactors**
  - User `__eq__`/`__hash__` now dispatch for every class; `==` calls user `__eq__` before identity; `!=` negates it.
  - Call machinery raises `TypeError: keywords must be strings` before any binder runs.
  - Dataclass metadata sits in the class namespace as `__dataclass_fields__`, so memory/GC and cycles are accounted via `Field` objects. Static strings fingerprint updated.
  - Dict insertion releases key/value when a memory limit rejects the entry, preventing leaked `Field`s if decoration aborts.

<sup>Written for commit 5ba8df55aebeb8b5c80f20e99f65578d2b9333d9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/626?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,585 @@
...
+    vm: &mut VM<'h>,
+    heap_ids: &mut LazyHeapSet,
+) -> RunResult<()> {
+    let class_id = instance.get(vm.heap).class();
+    let name = class_name(class_id, vm.heap, vm.interns).to_string();
+    // CPython's generated `__repr__` reads each field as an attribute, so an
+    // uninitialised one raises. Checked up front so no partial repr is emitted.
+    for name_id in field_names {
+        let field_name = vm.interns.get_str(*name_id).to_owned();
+        if instance
```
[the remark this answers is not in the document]
rewitt94: Valid, fixed in bc0674b8: `__repr__` and `__eq__` now resolve fields instance-dict-then-class-namespace via new `field_slot`/`field_value` helpers, so an unassigned defaulted field reads its class default while a field with no default still raises.

crates/monty/src/types/class.rs
```diff
@@ -32,13 +33,79 @@ pub(crate) struct Class {
...
+impl DataclassMeta {
+    /// Bytes this metadata adds to its `Class`, counting the field vector's
+    /// whole capacity.
+    ///
+    /// Decoration grows an already-allocated `Class` in place, so this must be
+    /// charged to the resource tracker explicitly — otherwise a sandbox could
+    /// decorate arbitrarily many wide classes without `max_memory` noticing.
+    #[must_use]
+    pub fn estimate_size(&self) -> usize {
+        mem::size_of::<Self>() + self.fields.capacity() * mem::size_of::<DataclassField>()
```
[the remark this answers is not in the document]
rewitt94: Correct, fixed in bc0674b8: `estimate_size` now returns only `self.fields.capacity() * size_of::<DataclassField>()`, since the inline `Option<DataclassMeta>` is already covered by `size_of::<Class>()`.

crates/monty/test_cases/refcount__dataclass_defaults.py
```diff
@@ -0,0 +1,26 @@
+# `@dataclass` captures each field's default at decoration time, so the metadata
+# on the `Class` holds an owned reference alongside the class namespace's own.
+#
+# After decoration `shared` is referenced by:
+#   - the module global `shared`
+#   - the class namespace entry `Defaults.b` (the annotated assignment)
+#   - the captured default in the class's dataclass metadata
+#
+# `nested` is only reachable through `shared`, so it keeps a single reference.
```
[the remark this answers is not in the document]
rewitt94: Right, fixed in bc0674b8: `nested` has 2 references (its own global plus `shared`'s tuple) and the decorator adds neither, since capturing a default clones the reference to `shared` itself.

crates/monty/src/types/instance.rs
```diff
@@ -565,6 +608,52 @@ fn instance_class(self_id: HeapId, vm: &VM<'_>) -> HeapId {
     }
 }
 
+/// Dispatches a user-defined `__eq__`, or `Ok(None)` when there is none, which
+/// leaves the caller on field-wise dataclass equality or identity.
+///
+/// The result is taken as a truth value: Monty has no `NotImplemented`, so a
+/// user `__eq__` cannot decline and the reflected operand is never tried.
+pub(crate) fn instance_user_eq(self_id: HeapId, other: &Value, vm: &mut VM<'_>) -> RunResult<Option<bool>> {
```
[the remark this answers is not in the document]
rewitt94: Already fixed in f9c4b6ac: `Value::py_eq_impl` dispatches `instance_user_eq` before the identity check (`crates/monty/src/value.rs:294-308`), leaving the container shortcut in `Value::py_eq`, and `class__eq_hash.py` asserts both (`not (never == never)` and `never in [never]`).

crates/monty/src/types/dataclass.rs
```diff
@@ -304,6 +280,45 @@ impl<'h> PyTrait<'h> for HeapRead<'h, Dataclass> {
...
+        if i > 0 {
+            f.write_str(", ")?;
+        }
+        let (field_name, value) = field(i, vm.heap, vm.interns);
+        f.write_str(&field_name)?;
+        f.write_char('=')?;
+        match value {
+            Some(value) => {
+                defer_drop!(value, vm);
+                value.py_repr_fmt(f, vm, heap_ids)?;
```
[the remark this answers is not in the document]
rewitt94: I believe this is not reproducible

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+/// Ownership of the single argument (the class) passes straight through to the
+/// return value, so no extra refcount work is needed on the happy path.
+fn dataclass_decorator(vm: &mut VM<'_>, args: ArgValues) -> RunResult<Value> {
+    // `@dataclass(...)`: name the unimplemented keyword form rather than letting
+    // the bare decorator's arity check report a confusing "0 given".
+    if matches!(args, ArgValues::Kwargs(_) | ArgValues::ArgsKargs { .. }) {
+        args.drop_with(vm);
+        return Err(ExcType::type_error(
+            "dataclass() keyword options (eq, order, frozen, unsafe_hash, ...) are not yet supported",
+        ));
```
davidhewitt: I think there's probably a `NotImplementedError` to use here.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+    }
+    let cls = args.get_one_arg("dataclass", vm.heap)?;
```
davidhewitt, line 84: I suggest having `defer_drop!(cls, vm);` here, then all the `drop_with` statements in each branch can go and instead `?` works in this function.

(The final line can do `cls.clone_with_heap(vm.heap)` to release it. Or if you really want to push for maximum efficiency, use a `DropGuard` instead of `defer_drop!` so you can reclaim the original value with `.into_inner()` in the final line.)

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+            return Err(SimpleException::new_msg(
+                ExcType::TypeError,
+                format!("dataclass() should be called on a class, not '{type_name}'"),
+            )
+            .into());
+        }
+    };
+    // Extract fields from `__annotations__`, then record the metadata. On a
+    // rejected class body the decorator's own reference must not leak.
+    let meta = match build_dataclass_meta(vm, class_id) {
```
davidhewitt: I think if a `DropGuard` is used on `meta` here (with a `DropWithContext` implementation on `DataclassMeta`) then a lot of the `drop_captured_defaults` stuff will be done on drop automatically and maybe `build_dataclass_meta` gets inlined into this function as 1-2 simple lines with `?`.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+
+/// Rejects a class body defining any of [`UNSUPPORTED_MEMBERS`].
+fn reject_unsupported_members(vm: &VM<'_>, class_id: HeapId) -> RunResult<()> {
+    match UNSUPPORTED_MEMBERS
+        .iter()
+        .find(|&&(name, _)| class_defines(class_id, name, vm))
+    {
+        Some((name, consequence)) => Err(ExcType::type_error(format!(
+            "dataclass() does not yet support {name} in a class body, {consequence}"
+        ))),
```
davidhewitt: Probably `NotImplementedError` again.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+/// checked — both divergences are in `limitations/dataclasses.md`.
+fn is_classvar(annotation: &str) -> bool {
+    annotation_head(annotation, "ClassVar")
+}
+
+/// Whether a stringized annotation denotes `dataclasses.InitVar`. Matched the
+/// same way as [`is_classvar`], and with the same textual-only caveats.
+fn is_initvar(annotation: &str) -> bool {
+    annotation_head(annotation, "InitVar")
+}
```
davidhewitt, line 293: This will do for now but we'll definitely want proper evaluation of annotations later.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+
+/// Constructs a dataclass instance by binding `args` to the recorded fields.
+///
+/// No bytecode is generated: arguments bind natively as CPython's synthesized
+/// `__init__(self, f1, f2=default, ...)` would, filling unbound fields from the
+/// defaults captured at decoration time. Drops the instance on error.
+pub(crate) fn dataclass_init(
+    vm: &mut VM<'_>,
+    class_id: HeapId,
+    instance_id: HeapId,
```
davidhewitt: This is a bit awkward, `class_id` is non-owning and `instance_id` is owning?

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+        Ok(values) => values,
+        Err(e) => {
+            Value::Ref(instance_id).drop_with(vm);
+            return Err(e);
+        }
+    };
+
+    // Store each bound value on the instance `__dict__`. On the rare set
+    // failure (resource limit) drop the untouched tail and the instance.
+    let mut values = values;
```
davidhewitt: This code looks like it's desperately wanting a `defer_drop_mut!` to simplify.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+    for i in 0..values.len() {
+        let value = mem::replace(&mut values[i], Value::None);
+        let name = Value::InternString(fields[i].0);
+        let set_result = if let HeapReadOutput::Instance(mut instance) = vm.heap.read(instance_id) {
```
davidhewitt: `vm.heap.read(instance_id)` should be done once before the loop.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+
+    if nonstring_key {
+        for v in positionals {
+            v.drop_with(vm);
+        }
+        for (_, v) in keywords {
+            v.drop_with(vm);
+        }
+        return Err(ExcType::type_error_kwargs_nonstring_key());
+    }
```
davidhewitt: This kind of `drop_with` branching makes me think both `positionals` and `keywords` probably can use `DropGuard` until they're consumed.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+        if !missing.is_empty() {
+            let refs: Vec<&str> = missing.iter().map(String::as_str).collect();
+            error = Some(ExcType::type_error_missing_positional_with_names(&init_name, &refs));
+        }
+    }
+
+    // Excess positionals are never bound, so release them on every path.
+    for v in excess {
+        v.drop_with(vm);
+    }
```
davidhewitt: I'm pretty sure this can be


```suggestion
    excess.drop_with(vm);
```

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+
+    if let Some(e) = error {
+        for v in values.into_iter().flatten() {
+            v.drop_with(vm);
+        }
+        return Err(e);
+    }
+    Ok(values.into_iter().map(|v| v.expect("all fields bound")).collect())
```
davidhewitt: If `values` and `excess` are in `DropGuard`s the error returns can probably go early and the branches simplify here.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,632 @@
...
+        let field_name = vm.interns.get_str(*name_id).to_owned();
+        let a = instance_attr(self_id, &field_name, vm)?;
+        // A failure here (binding a method can allocate) must not strand `a`.
+        let b = match instance_attr(other_id, &field_name, vm) {
+            Ok(b) => b,
+            Err(e) => {
+                a.drop_with(vm);
+                return Err(e);
+            }
+        };
```
davidhewitt:
```suggestion
        let field_name = vm.interns.get_str(*name_id).to_owned();
        let a = instance_attr(self_id, &field_name, vm)?;
        defer_drop!(a, vm);
        let b = instance_attr(other_id, &field_name, vm)?;
        defer_drop!(b, vm);
```

... and once the deferred drops are in it should be the case that the branches below simplify too.

crates/monty/src/types/dataclass.rs
```diff
@@ -304,6 +280,50 @@ impl<'h> PyTrait<'h> for HeapRead<'h, Dataclass> {
...
+                break;
+            }
+            f.write_str(", ")?;
+        }
+        let (field_name, value) = field(i, vm.heap, vm.interns);
+        f.write_str(&field_name)?;
+        f.write_char('=')?;
+        match value {
+            Some(value) => {
+                defer_drop!(value, vm);
```
davidhewitt: I think the `defer_drop!(value, vm)` needs to go before the `write_str` calls on line 314 else it leaks on error returns.

crates/monty/src/modules/dataclasses.rs
rewitt94: Swapped for `NotImplementedError`

crates/monty/src/modules/dataclasses.rs
rewitt94: Improved how we handle ref decrementing file wide

crates/monty/src/modules/dataclasses.rs
rewitt94: Mmmm this is awkward - should instance_id become value as that signifies owning? But then the enum becomes generic. Do you have a suggestion?

davidhewitt: I would probably take it as `Value` yes, and maybe do `vm.heap.read(instance_id)` as the first line inside this function and use `&HeapRead<'h, Instance>` as the thing passed to inner functions.

Similarly it looks like the caller of this function in `vm/call.rs` has already done a read on class, so knows it's a class, probably `class_id` should be `&HeapRead<'h, Class>` and then it's very obvious it's not transferring ownership.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,572 @@
...
+    }
+    let cls = args.get_one_arg("dataclass", vm.heap)?;
+    // The decorator returns the class it was given, so `cls` is only released
+    // when something below rejects it — which the guard handles on every path.
+    let mut guard = DropGuard::new(cls, vm);
+    let (cls, vm) = guard.as_parts();
+    let class_id = match cls {
+        Value::Ref(id) if matches!(vm.heap.get(*id), HeapData::Class(_)) => *id,
+        _ => return Err(non_class_error(cls, vm)),
+    };
```
davidhewitt: I have a feeling that this should be


```suggestion
    let HeapReadOutput(mut class) = vm.heap.get(*id) else {
        return Err(non_class_error(cls, vm));
    };
```

... and from here down (including in functions called from here) `class_id` should be replaced with `HeapRead<'_, Class>`.

I think that'll remove a lot of pointless matches on `heap.read(class_id)` and `heap.get(class_id)`.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,572 @@
...
+            (key, value).drop_with(vm);
+            nonstring_key = true;
+            continue;
+        };
+        key.drop_with(vm);
+        // Past the first error the mapping is only drained, never bound.
+        if kw_error.is_some() {
+            value.drop_with(vm);
+            continue;
+        }
```
davidhewitt: It looks like here if you do

```rust
let kwargs_iter = kwargs.into_iter();
defer_drop_mut!(kwargs_iter, vm);
```

then you could just have a loop like

```rust
for (key, value) in kwargs_iter {
    defer_drop!(key, vm);
    let value_gard = DropGuard::new(value, vm);
    // do stuff
       Some(idx) if values[idx].is_none() => values[idx] = Some(value_guard.into_inner()),
    // ...
}
```

where `do stuff` can early return freely and no need to have `nonstring_key` or `kw_error` state.

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,572 @@
...
+    field_names: &[StringId],
+    other: &Value,
+    vm: &mut VM<'_>,
+) -> RunResult<Option<bool>> {
+    let class_id = instance_class(self_id, vm);
+    // Only the same dataclass type can be equal; otherwise NotImplemented.
+    let Value::Ref(other_id) = other else {
+        return Ok(None);
+    };
+    let other_id = *other_id;
```
davidhewitt: I think this might work:


```suggestion
    let &Value::Ref(other_id) = other else {
        return Ok(None);
    };
```

crates/monty/src/types/class.rs
```diff
@@ -32,13 +35,98 @@ pub(crate) struct Class {
...
+/// Metadata recorded on a [`Class`] by the `@dataclass` decorator.
+///
+/// **Holds owned heap references** through its captured defaults, so a `Class`
+/// carrying one counts them in `py_estimate_size`, `py_dec_ref_ids` (both below)
+/// and `heap::for_each_child_id`.
+#[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
+pub(crate) struct DataclassMeta {
+    /// Fields in definition order (from `__annotations__`, `ClassVar` excluded).
+    pub fields: Vec<DataclassField>,
+}
```
davidhewitt: Rather than having this `dataclass_meta` in the `Class` struct, I think we should store this in the `namespace` as a child dict:

```python
>>> import dataclasses
>>> @dataclasses.dataclass
... class Foo:
...     x: int
...
>>> Foo.__dataclass_fields__
{'x': Field(name='x',type=<class 'int'>,default=<dataclasses._MISSING_TYPE object at 0x70ab73b746e0>,default_factory=<dataclasses._MISSING_TYPE object at 0x70ab73b746e0>,init=True,repr=True,hash=None,compare=True,metadata=mappingproxy({}),kw_only=False,doc=None,_field_type=_FIELD)}
```

... we'll need `Field` anyway as part of the full dataclasses API

crates/monty/src/types/class.rs
```diff
@@ -32,13 +35,98 @@ pub(crate) struct Class {
...
+    pub name: StringId,
+    /// The default **captured when `@dataclass` ran** (an owned reference), or
+    /// `None` for a required field. CPython bakes defaults into the generated
+    /// `__init__`, so rebinding the class attribute later must not change it.
+    pub default: Option<Value>,
+    /// Whether the annotation was `InitVar[...]`. Recorded but not acted on —
+    /// decoration rejects such a field until pseudo-fields are implemented.
+    #[serde(default)]
+    pub initvar: bool,
 }
```
davidhewitt: And with `DataclassMeta` stored as `__dataclass_fields__` in the namespace, this can be a new `HeapData` type and belong in `dataclasses.rs` (or a child module).

crates/monty/src/types/class.rs
davidhewitt, line 1: I think with the suggestions below to move `DataclassMeta` and `DataclassField` we can avoid changing this file at all

davidhewitt (commented):
Generally looking good, I'd like to see a few more tidy-ups (hopefully leading to simplifications) and then I think this'll be ready to merge!

crates/monty/src/types/class.rs
rewitt94: Have removed dataclass_meta within `Class` and made a `HeapData::DataclassField`.

crates/monty/src/types/class.rs
rewitt94: Have removed dataclass_meta within `Class` and made a `HeapData::DataclassField`.

crates/monty/src/types/class.rs
rewitt94: Done

crates/monty/src/modules/dataclasses.rs
rewitt94: Done

crates/monty/src/modules/dataclasses.rs
rewitt94: Done

crates/monty/src/modules/dataclasses.rs
rewitt94: Did more or less this

crates/monty/src/modules/dataclasses.rs
```diff
@@ -0,0 +1,641 @@
...
+    // Up front, because the call machinery raises this before `__init__` binds
+    // anything: it outranks every error below whatever its position among the
+    // keys, and checking it here leaves the binding loop free to raise on the
+    // spot (see `KwargsValues::has_nonstring_key`).
+    if let ArgValues::Kwargs(kwargs) | ArgValues::ArgsKargs { kwargs, .. } = &args
+        && kwargs.has_nonstring_key(vm.heap)
+    {
+        args.drop_with(vm);
+        return Err(ExcType::type_error_kwargs_nonstring_key());
+    }
```
davidhewitt: Local AI analysis tells me this is redundant because the check for string keys is done in `dict_merge_inner` in `vm/collections.rs`, perhaps remove?

crates/monty/src/args/mod.rs
```diff
@@ -337,6 +337,22 @@ impl KwargsValues {
...
+    /// front for `f(**{'z': 1, 1: 2})` to report the non-string key rather than
+    /// the unexpected `'z'` it would reach first.
+    #[must_use]
+    pub fn has_nonstring_key(&self, heap: &Heap) -> bool {
+        match self {
+            Self::Empty | Self::Inline(_) => false,
+            Self::Pairs(kvs) => kvs.iter().any(|(key, _)| key.as_either_str(heap).is_none()),
+            Self::Dict(dict) => dict.iter().any(|(key, _)| key.as_either_str(heap).is_none()),
+        }
+    }
```
davidhewitt: Probably don't need this based on the other comment about the check being redundant.

rewitt94: Removed, thanks

crates/monty/src/modules/dataclasses.rs
rewitt94: Removed, thanks

davidhewitt (approved):
Very nice 👍

PR #650, opened 2026-08-03
Support `@dataclass(eq=..., frozen=...)`

rewitt94:
Adds the `@dataclass(...)` keyword form, for the two options Monty implements: **`eq`** and **`frozen`**.

**The decorator is a value.** `@dataclass(frozen=True)` is two calls, so the options must live in a callable between them. Monty's callables were all stateless, so `ModuleFunction` gained a variant that carries them — cheap, because they are two `Copy` bools. A runtime value rather than a compile-time form, since the decorator can be stored and reused.

**Options live on the class, fields in the namespace.** That is CPython's split: it bakes the options into the dunders it generates, and reads `__dataclass_fields__` live. So `__dataclass_params__` is written as a report and never consulted, and the instance write path stays free of lookups.

**The dunders stay native.** `frozen` gates `Instance::set_attr`; `eq`/`frozen` choose between a synthesized field-wise `__eq__`/`__hash__` and identity. No Python-level dunder dispatch and no generated code. The field-wise hash charges a recursion level, as equality does, since a chain of frozen dataclasses nests as deep as it is built.

**Unimplemented options are refused by name**, not ignored — `FromArgs` binds CPython's full signature, so each of the other eight reports itself.

Dump `VERSION` moves 3 → 4: `HeapData` gains a variant and `Class` a field.

Divergences are documented in `limitations/dataclasses.md`. 1122 test cases dual-run against CPython 3.14; rust suites, `ref-count-return`, clippy and `lint-py` green.

---
## Summary by cubic
Adds CPython-aligned support for `@dataclass(eq=..., frozen=...)`, including the keyword form and a reusable configured decorator. This improves correctness (equality, hashing, and freezing) and matches CPython’s signatures and errors.

- Decorator: accepts `eq` and `frozen` by truthiness; returns a configured callable (`dataclass.<locals>.wrap`) that accepts `cls` positionally or by keyword; arity/keyword errors match CPython.
- Frozen: blocks any instance attribute write with `dataclasses.FrozenInstanceError` and message "cannot assign to field '<name>'"; synthesized `__init__` writes through an unchecked path.
- Equality/Hashing: `eq=True` synthesizes field-wise `__eq__`; `eq=False` uses identity and stays hashable. Hashing follows CPython’s table (incl. “has explicit hash”): `frozen=True` + `eq=True` hashes by field values; `eq=True` + mutable is unhashable; an explicit `__hash__` is respected; `__hash__ = None` next to a body `__eq__` is overwritten. Recursion is guarded.
- Introspection: writes `__dataclass_params__` (type `_DataclassParams`) and records options on the class; all ten flags read as booleans. Unimplemented options (`init`, `repr`, `order`, `unsafe_hash`, `match_args`, `kw_only`, `slots`, `weakref_slot`) are rejected by name when set away from defaults. Exposes `dataclasses.FrozenInstanceError`.

- Migration
  - Dump format bumped to version 4; regenerate persisted dumps.

<sup>Written for commit a6f6f6fa3cdd4b39670ec1ddf5f5aad800156299. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/650?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/src/modules/dataclasses/options.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    /// CPython's `_DataclassParams.__repr__`, flag for flag. The eight
+    /// unimplemented options print as constants, which cannot misreport a class
+    /// because any other value is refused at decoration.
+    fn py_repr_fmt(&self, f: &mut impl Write, vm: &mut VM<'h>, _heap_ids: &mut LazyHeapSet) -> RunResult<()> {
+        let options = self.get(vm.heap).options;
+        Ok(write!(
+            f,
+            "_DataclassParams(init=True,repr=True,eq={},order=False,unsafe_hash=False,frozen={},\
```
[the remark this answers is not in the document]
rewitt94: Cubic is comparing to an old python - this is not true on 3.14 which monty supports

crates/monty/src/modules/dataclasses.rs
```diff
@@ -95,7 +206,13 @@ fn dataclass_decorator(vm: &mut VM<'_>, args: ArgValues) -> RunResult<Value> {
         return Err(non_class_error(cls, vm));
     };
     let fields = build_dataclass_fields(&class, vm)?;
+    // Fields first, so nothing owns the mapping while the params allocation
+    // can still fail; a class left with fields alone reads back as a default one.
     store_dataclass_fields(&mut class, fields, vm)?;
+    store_dataclass_params(&mut class, options, vm)?;
+    // Last, so the options a class acts on are only ever those of a decoration
+    // that ran to completion.
+    class.set_dataclass_options(options, vm);
```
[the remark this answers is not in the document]
rewitt94: Fixed in https://github.com/pydantic/monty/pull/650/changes/fee7ab7e8797736a8513cb22d2ddf2aeb29a36bd

crates/monty/src/types/class.rs
```diff
@@ -32,13 +33,35 @@ pub(crate) struct Class {
     name: EitherStr,
     /// Members: method name / class-variable name -> value.
     namespace: Dict,
+    /// The `@dataclass(...)` options this class was decorated with, left at
+    /// CPython's defaults for a class that was not. Baked in here at decoration
+    /// rather than read back from `__dataclass_params__`, because CPython bakes
+    /// them into the dunders it generates and never consults that object again.
+    options: DataclassOptions,
```
davidhewitt, line 64: I'm not sure I understand why we store this here instead of reading the attribute from the class namespace?

crates/monty/src/types/instance.rs
```diff
@@ -72,7 +72,23 @@ impl<'h> HeapRead<'h, Instance> {
...
     /// caller to drop. Takes ownership of both `name` and `value`.
+    ///
+    /// The one hook a `@dataclass(frozen=True)` needs on the write path; what
+    /// counts as frozen and how it reads is [`dataclasses`]' business.
     pub fn set_attr(&mut self, name: Value, value: Value, vm: &mut VM<'h>) -> RunResult<Option<Value>> {
+        let class_id = self.get(vm.heap).class();
+        if let Some(exc) = dataclasses::frozen_assignment_error(class_id, &name, vm) {
+            [name, value].drop_with(vm);
+            return Err(exc);
+        }
```
davidhewitt, line 83: It seems unfortunate that `set_attr` on user-defined types gains knowledge of dataclass code.

I wonder if we can implement `__setattr__` dispatch in user-defined code and then somehow install a dataclass-generated `__setattr__` for frozen types so that non-dataclass types pay no price here?

davidhewitt (commented):
Generally looks great, thanks. Just a couple of structural ramblings that might be best left for follow-ups...

crates/monty/src/types/instance.rs
rewitt94: Sounds good I will start another PR for looking at `__setattr__` - I think out of scope here though

crates/monty/src/types/class.rs
rewitt94: The namespace i.e. `__dataclass_params__` is user-rewritable.
CPython would generate dunders, storing options here seems OK?
I will move DataclassOptions into `types` not `modules` so the dependency seems less out of place.

davidhewitt (approved):

PR #654, opened 2026-08-04
Dispatch `__index__` for indexing, slicing and integer arguments

rewitt94:
`__index__` is Python's "this object *is* an integer" protocol (PEP 357), called wherever an exact integer is required for a non-arithmetic purpose: sequence subscripts, slice bounds, `range()`, and integer arguments like `'x'.center(n)`. It is deliberately narrower than `__int__` — a lossy conversion that `float` also defines — which is why `seq[3.7]` is a `TypeError` while a class defining `__index__` indexes fine.

**The gap.** Raised by cubic on [pydantic/monty#635](https://github.com/pydantic/monty/pull/635) as an `islice` bug, but not `islice`-specific: Monty dispatched ten dunders, not `__index__`, so a class defining it was rejected everywhere CPython accepts an index, though the error messages already promised the protocol and `monty -t` already accepted the code.

**The fix.** `instance_index` joins the existing `instance_call_dunder_sync` dispatch and validates an int result; `as_int`/`as_index` now take `&mut VM`. Four hand-rolled int fast paths fold into them, picking up `LongInt` and correcting two `str` argument errors to match CPython.

**Not covered.** Sequence repetition (`'ab' * obj`) and `slice()` storing coerced bounds rather than the original object — both documented in `limitations/classes.md`.

**Testing.** New `class__index_dunder.py` / `refcount__index_dunder.py` cases dual-run against CPython; suite green under `ref-count-return` and `memory-model-checks`.

**Unrelated, pre-existing.** An exception raised in any sync dunder escapes `try`/`except` when the receiver is a variable (`1 in c`, but not `1 in C()`). Reproduces on clean `main`; not fixed here.

### DIFF
--- a/crates/monty/src/bytecode/vm/mod.rs
+++ b/crates/monty/src/bytecode/vm/mod.rs
@@ -78,6 +78,17 @@ macro_rules! try_catch_sync {
     };
 }
 
+/// Persists the cached frame IP before an operation that can execute Python.
+///
+/// A nested `run()` restores `instruction_ip` from the caller frame, so the
+/// caller must hold its resume IP for exceptions to find the right handler.
+macro_rules! try_catch_reentrant {
+    ($self:expr, $cached_frame:ident, $expr:expr) => {{
+        $self.current_frame_mut().ip = $cached_frame.ip;
+        try_catch_sync!($self, $cached_frame, $expr);
+    }};
+}
+
 /// Handles an exception and reloads cached frame state if caught.
 ///
 /// Use this in the main run loop where `cached_frame`
@@ -1317,7 +1328,8 @@ impl<'h> VM<'h> {
                     try_catch_sync!(self, cached_frame, self.build_fstring(count));
                 }
                 Opcode::BuildSlice => {
-                    try_catch_sync!(self, cached_frame, self.build_slice());
+                    // Bounds are coerced here, so a user `__index__` can run.
+                    try_catch_reentrant!(self, cached_frame, self.build_slice());
                 }
                 Opcode::ListExtend => {
                     try_catch_sync!(self, cached_frame, self.list_extend());
@@ -1359,6 +1371,9 @@ impl<'h> VM<'h> {
                 Opcode::BinarySubscr => {
                     let index = self.pop();
                     let obj = self.pop();
+                    // Sync IP before the call: a user `__index__` runs a nested
+                    // run() loop, which restores the frame IP from this frame.
+                    self.current_frame_mut().ip = cached_frame.ip;
                     let result = obj.py_getitem(&index, self);
                     obj.drop_with(self);
                     index.drop_with(self);
@@ -1512,9 +1527,13 @@ impl<'h> VM<'h> {
                     let (type_id, arg_count) = cached_frame.fetch_u8_u8();
                     let arg_count = arg_count as usize;
 
+                    // Sync IP before the call: `range(obj)`/`slice(obj)` coerce
+                    // their arguments, so a user `__index__` can push a frame
+                    // and run a nested run() loop.
+                    self.current_frame_mut().ip = cached_frame.ip;
+
                     match self.exec_call_builtin_type(type_id, arg_count) {
                         Ok(result) => self.push(result),
-                        // IP sync deferred to error path (no frame push possible)
                         Err(err) => catch_sync!(self, cached_frame, err),
                     }
                 }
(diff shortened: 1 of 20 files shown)


crates/monty/test_cases/class__index_dunder.py
```diff
@@ -0,0 +1,219 @@
...
+            end_target.pop()
+        return -1
+
+
+# The window becomes [0, 3), so the 4 left at index 3 is out of range.
+try:
+    end_target.index(4, 0, ShrinkEnd())
+    assert False, 'expected the stale end bound to exclude the match'
+except ValueError as exc:
+    assert str(exc) == 'list.index(x): x not in list'
```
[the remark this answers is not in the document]
rewitt94: `ValueError: '4 is not in list'` -> reproduced on CPython 3.9.6

`ValueError: 'list.index(x): x not in list'` -> actual error on CPython 3.14.6

This is invalid for the version of python we target

rewitt94:
Fwiw, I cannot reproduce locally that this branch is slower than main. Can we re-run the code speed check here?

crates/monty/src/bytecode/vm/mod.rs
```diff
@@ -78,6 +78,17 @@ macro_rules! try_catch_sync {
...
+/// Persists the cached frame IP before an operation that can execute Python.
+///
+/// A nested `run()` restores `instruction_ip` from the caller frame, so the
+/// caller must hold its resume IP for exceptions to find the right handler.
+macro_rules! try_catch_reentrant {
+    ($self:expr, $cached_frame:ident, $expr:expr) => {{
+        $self.current_frame_mut().ip = $cached_frame.ip;
+        try_catch_sync!($self, $cached_frame, $expr);
+    }};
+}
```
davidhewitt, line 90: I think #762 will remove the need for this 🎉

crates/monty/src/types/instance.rs
```diff
@@ -510,6 +510,36 @@ pub(crate) fn instance_str(self_id: HeapId, vm: &mut VM<'_>) -> RunResult<Value>
...
+///
+/// The returned value is validated to be a real `int` (`Int`/`Bool`/`LongInt`),
+/// which is both CPython's `__index__ returned non-int` check and what stops a
+/// class whose `__index__` returns another such instance from recursing: the
+/// caller can convert the result without re-entering this path. Out-of-range
+/// policy stays with the caller, since `as_int` and `as_index` differ on it.
+///
+/// Runs synchronously like the other dunders here, so an `__index__` that calls
+/// an external/OS function raises rather than suspending.
+pub(crate) fn instance_index(self_id: HeapId, vm: &mut VM<'_>) -> RunResult<Option<Value>> {
```
davidhewitt: This should probably use `HeapObjectRead<'h, Instance>` as the first argument (or even be a method on that type).

crates/monty/src/value.rs
```diff
@@ -1771,56 +1767,128 @@ impl Value {
...
+    /// than testing for `HeapData::Instance` themselves, so how a user class is
+    /// represented on the heap stays known only to this module and `instance.rs`.
+    /// The result is a validated int, so a caller may convert it — or recurse
+    /// once through its own int arms — without re-entering the dunder.
+    pub(crate) fn try_index(&self, vm: &mut VM<'_>) -> RunResult<Option<Self>> {
+        match self {
+            Self::Ref(id) if matches!(vm.heap.get(*id), HeapData::Instance(_)) => instance_index(*id, vm),
+            _ => Ok(None),
+        }
+    }
```
davidhewitt: This feels to me like this probably should be a `PyTrait::py_index` method.

crates/monty/src/bytecode/vm/mod.rs
rewitt94: Oo cool, which shall we merge first?

crates/monty/src/types/str.rs
```diff
@@ -1199,11 +1206,33 @@ fn optional_index(value: &Value, default: usize, str_len: usize, vm: &mut VM<'_>
...
+        // error, this keeps an interned one out of the `_` arm below, whose
+        // recursion would otherwise hand it straight back to itself.
+        Value::InternLongInt(id) => {
+            let i = vm
+                .interns
+                .get_long_int(*id)
+                .to_i64()
+                .ok_or_else(|| ExcType::type_error("integer too large"))?;
+            Ok(normalize_sequence_index(i, str_len))
+        }
```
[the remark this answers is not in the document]
rewitt94: This is valid, but the issue is fairly widespread I will raise a follow up PR instead of fixing here.

davidhewitt (approved):
Let's merge this first, I'll merge #762 after.

PR #655, opened 2026-08-04
Windows path-parsing mount escape

rewitt94:
On Windows, `PathBuf::join` discards the mount base when a `/`-split virtual-path segment carries a drive/UNC/root prefix (`C:\…`, `C:`, `\\host\share`), so an untrusted path escapes the mount. Fixed by rejecting such segments in `ResolutionRequest::new` before the join, on all hosts.

Three variants, one root cause:

- **Out-of-mount write** — `os.makedirs(parents=True)` creates a directory outside the mount.
- **Existence oracle** — `os.path.exists` leaks whether any host path exists (`PathEscape` vs `False`).
- **UNC egress** — resolving `\\host\share\x` triggers an SMB/NTLM handshake to an attacker host.

**CI:** the repro is one parameterized test gated `#[cfg_attr(not(windows), ignore)]`, so it runs only on the Windows leg of `cargo test -p monty-fs`.

Based on VERIA-3.

---
## Summary by cubic
Blocks Windows mount escapes by rejecting drive/UNC/root-like segments before any join and by boundary-checking joined/returned paths across all hosts and modes (including overlay keys). Closes VERIA-3 by stopping out-of-mount writes, existence-oracle leaks, and UNC egress.

- **Bug Fixes**
  - Enforced `reject_drive_or_unc_segments` in `ResolutionRequest::new` and `overlay::relative_path`; refuse `C:\…`, `C:`, `\\host\share`, and any backslash-containing segment on every OS/mode.
  - Added `check_boundary` on the just-joined candidate before any host I/O, and on paths returned by `resolve_mkdir_parents`; traps base-clobbering segments early and ensures no mode returns an unchecked path.
  - Tests run on all hosts and modes: refusal checked before I/O for all ops; POSIX-absolute segments are confined inside the mount; `exists` is not an oracle yet still discriminates in-mount; `OverlayMemory` rejects these as keys; colon handling pinned (only `<letter>:` is rejected; other colons allowed).
  - Docs updated to state confinement for POSIX-absolute segments and uniform `<letter>:` rejection; other colons are fine.
  - Lint: outcome classifier now takes `Option<&T>` to satisfy `clippy::ref_option`.

- **Refactors**
  - Trimmed comments/docstrings and reworded a note to satisfy `codespell`; no behavior change.

<sup>Written for commit 88b3c8649dc2ac733abf45575eed2779b333228f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/655?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty-fs/src/path_security.rs
```diff
@@ -301,6 +302,33 @@ pub(super) fn reject_overlong_path(normalized: &str, original: &str) -> Result<(
...
 
+/// Rejects segments a host parser treats as drive/UNC/root-absolute: on
+/// Windows `PathBuf::join` discards the mount base for `C:\x`, `C:`, or
+/// `\\host\share`, escaping before any boundary check. Runs on all hosts for
+/// consistent behavior, and must precede the `candidate_host` join.
+///
+/// `OverlayMemory` needs it too: it serves requests from memory via
+/// `overlay::relative_path`, where these segments would otherwise become valid
+/// keys that every other mode refuses.
+pub(super) fn reject_drive_or_unc_segments(relative: &str, normalized_virtual_path: &str) -> Result<(), MountError> {
```
samuelcolvin, line 320: should we `cfg` this to only run on windows?

crates/monty-fs/src/path_security.rs
```diff
@@ -130,6 +130,7 @@ impl ResolutionRequest {
         let relative = strip_mount_prefix(&normalized_virtual, mount_virtual_path)
             .ok_or_else(|| MountError::NoMountPoint(virtual_path.to_owned()))?
             .to_owned();
+        reject_drive_or_unc_segments(&relative, &normalized_virtual)?;
```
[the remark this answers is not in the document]
samuelcolvin: I'm not sure this makes sense, but if it does, it needs to be addressed.

samuelcolvin (commented):
@rewitt94 did you actually try this on windows? If not, I think it's worth doing so, or waiting for @davidhewitt if you don't have a windows box.

rewitt94:
@samuelcolvin - yes I did try this on my windows! The reprodution test cases failed on my windows machine

crates/monty-fs/src/path_security.rs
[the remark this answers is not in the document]
rewitt94: It did make sense! There is a commit to fix it already

`b1fddd88` — "Apply the host-absolute segment check to OverlayMemory, test on all hosts"

crates/monty-fs/src/path_security.rs
rewitt94: Not sure. Maybe it's better to be unconditional so that monty will allow the same file paths (be consistent) regardless of the OS the sandbox is running on. 🤔

rewitt94: We agreed not to cfg to only windows so we are consistent with filepaths regardless of OS

PR #657, opened 2026-08-04
Add `itertools` adaptors: `takewhile`, `dropwhile`, `filterfalse`, `starmap`

rewitt94:
Four more `itertools` callables on the source-wrapping foundation from #635 (now merged, so this is rebased onto `main`): `takewhile`, `dropwhile`, `filterfalse` and `starmap`. All four apply a **user callable per item**, which `pairwise`/`compress`/`islice`/`chain`/`cycle` do not — that is the new capability and the main thing to review.

**Applying the callable re-enters the VM.** `next` calls `evaluate_function`, which can raise or allocate, so the item under test is held in a `DropGuard` across the call: it is yielded on one answer and dropped on the other. The callable is also a **second GC edge** alongside the source, so `for_each_child_id` and `py_dec_ref_ids` trace both.

**Latching differs per adaptor and is user-visible.** `takewhile` latches `done` on the first rejection, so neither the predicate nor the source is touched again. `dropwhile` clears `dropping` on the first failure and yields everything after it untested. `filterfalse` and `starmap` have no spent flag at all — source exhaustion is their only end condition.

**`filterfalse` accepts `None`.** It shares the one-argument application in `predicate.rs` with the two `while` adaptors, plus the truth-test path `filter(None, ...)` uses.

**Spreading is arity-sensitive.** `starmap` packs each item into `ArgValues::Empty`/`One`/`Two` rather than `ArgsKargs`, because extractors such as `get_one_arg` match the shape structurally — a one-element `ArgsKargs` is rejected as the wrong shape (`abs() takes exactly one argument`) even though the arity is right.

**A callable that suspends is rejected, not paused.** `evaluate_function` runs a frame to completion and cannot yield to the host, so a callable reaching an external function or an `os` operation raises `NotImplementedError` where CPython would simply call it — the same restriction as `__init__`/`__next__`/`__repr__`. Documented in `limitations/itertools.md` and covered by a Rust-side test.

**Heap footprint.** The four new variants are 32–40 bytes, under `Islice`/`Chain`'s 56, so `ItertoolsIter` stays at 64 and `HeapData` at 72 — inside the 80-byte ceiling from #636.

Not implemented: `accumulate`, `batched`, `combinations`, `combinations_with_replacement`, `groupby`, `permutations`, `product`, `tee`, `zip_longest`.

Compatible dump change: new `StaticStrings`, `Type`, `MontyType` and `ItertoolsIter` variants are appended, so only the fingerprint moves.

---
## Summary by cubic
Adds `itertools` adaptors `takewhile`, `dropwhile`, `filterfalse`, and `starmap`, and closes `max_duration` gaps in native adaptor loops and eager drains. Also classifies these as concrete iterators so `__iter__` can return them.

- Per-item callables: executed via `evaluate_function`; external/host callables raise `NotImplementedError` naming the adaptor. `filterfalse(None, iterable)` keeps falsy items. `starmap` spreads each item into arguments; a non-iterable item raises `TypeError`. All four take exactly two positional-only args; keywords are rejected; the iterable resolves eagerly (non-iterables raise up front). Typeshed stubs are updated.
- Semantics: `takewhile` latches on the first predicate failure, releases its predicate and source, and does not latch on transient `StopIteration`. `dropwhile` stops dropping after the first failure, yields the rest untested, and retains its predicate until destruction. `filterfalse` and `starmap` never latch and re-drive past transient `StopIteration`.
- Time limits: native discard loops in `dropwhile`/`filterfalse` and in `chain`/`compress`/`islice` poll the tracker; eager drains, including `starmap` and the shared iterator drain, also check time per yield. Overhead is amortized; behavior otherwise unchanged.
- Internal: shared per-item step logic for predicate-driven adaptors; new `predicate::call_predicate` used by `filter()`. New `Type`/`MontyType`/`StaticStrings`/`ItertoolsIter` variants; adaptors treated as iterator types; size-budget assertion gated to 64‑bit hosts; dump fingerprint bumped.

<sup>Written for commit 8d02c3593a5a2bbc25ad4a22da14950a9d5fc561. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/657?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/src/predicate.rs
```diff
@@ -0,0 +1,21 @@
...
+
+/// Applies `predicate` to `item` and returns its truthiness.
+///
+/// `item` is only borrowed — the call gets its own reference, so the caller can
+/// still yield the item afterwards. Goes through `evaluate_function`, which
+/// runs a defined function's frame to completion, so a predicate that suspends
+/// (an external function, an `os` call) is rejected rather than paused; `ctx`
+/// names the adaptor in that error.
+pub(super) fn call_predicate(predicate: &Value, item: &Value, ctx: &'static str, vm: &mut VM<'_>) -> RunResult<bool> {
+    let arg = item.clone_with_heap(vm.heap);
```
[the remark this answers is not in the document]
rewitt94: Done — `builtin_filter` now calls the shared helper. `predicate.rs` moves to the crate root (`crate::predicate`) and `call_predicate` becomes `pub(crate)`, so `builtins/` does not reach into `types::itertools::`.

crates/monty/src/builtins/filter.rs
```diff
@@ -48,12 +49,7 @@ pub fn builtin_filter(vm: &mut VM<'_>, args: ArgValues) -> RunResult<Value> {
             // No predicate - use truthiness of element
             item.py_bool(vm)?
         } else {
-            // Clone for predicate call - the clone is consumed by evaluate_function
-            let item_for_predicate = item.clone_with_heap(vm);
-            let result = vm.evaluate_function("filter()", function, ArgValues::One(item_for_predicate))?;
-            let is_truthy = result.py_bool(vm);
-            result.drop_with(vm);
-            is_truthy?
+            call_predicate(function, item, "filter()", vm)?
```
rewitt94, line 52: This diff is here because cubic asked for it (see the thread on `predicate.rs`). `builtin_filter` is pre-existing and otherwise untouched by this branch — the change routes it through `call_predicate`, the helper this PR adds for `takewhile`/`dropwhile`/`filterfalse`, so the two predicate-call paths cannot drift apart. Behaviour-neutral.

crates/monty/src/types/itertools/mod.rs
```diff
@@ -51,8 +63,21 @@ pub(crate) enum ItertoolsIter {
...
 }
 
+// `Dict` is the widest `HeapData` payload on 64-bit hosts, so it — not a
+// literal — is the budget: staying under it keeps this family from setting
+// `HeapData`'s size. Only there: on 32-bit (the wasm worker) `Dict` halves
+// while the adaptors' `i64` fields do not, and other variants set the size.
+// TODO: when this fails, box the offending variant (`GroupBy(Box<GroupBy>)`),
+// not the enum and not at the `HeapData` boundary.
+#[cfg(target_pointer_width = "64")]
+const _: () = assert!(mem::size_of::<ItertoolsIter>() <= mem::size_of::<Dict>());
```
rewitt94, line 80: We will need to box ItertoolsIter soon as per https://github.com/pydantic/monty/pull/636 - this is just a compile check to remind

rewitt94:
Close and open to re-trigger crashed CI

davidhewitt (approved):

davidhewitt (approved):

Issue #698, opened 2026-08-09
Announcing a suspension costs extra copies of its payload, so a large host-call argument can exceed max_memory


samuelcolvin:
_(written by claude)_

## Summary

Handing a value **to** the host costs more transient copies of it than returning
the same value does, so a host-call argument that is a large fraction of
`max_memory` can kill the worker even though the value itself fits the budget
several times over.

Originally found against `monty-server` (private) at rev `dd6139a`, but nothing
about it is server-specific — it is the child's own serialisation path, so every
`monty-pool` embedder pays it.

## Reproduction

Session limits: `max_memory = 10 MiB` (allocator hard ceiling is
`baseline + 10 MiB + 4 MiB` of `BASE_HEADROOM`). Fed code:

```python
s = "A" * (4 * 1024 * 1024 - 10)
result = foobar(s)          # foobar is a host function
result
```

The worker dies mid-announcement, having logged:

```
monty worker: allocation of 4194327 bytes exceeds the memory limit
```

That is the string (`4194294`) plus 33 bytes of protobuf framing — i.e. the
allocation that crossed the ceiling is the *encode buffer* for the suspension
frame, at a point where ~10 MiB was already live for a 4 MiB string.

Probing the same session isolates it to the suspension path:

| snippet | outcome |
| --- | --- |
| `s = "A" * (8 * 1024 * 1024); len(s)` | ok — 8 MiB allocation is fine |
| `s = "A" * (4 * 1024 * 1024 - 10); s` | ok — 4 MiB *returned* to the host |
| `foobar("A" * (2 * 1024 * 1024))` | ok |
| `foobar("A" * (4 * 1024 * 1024 - 10))` | **worker killed** |

So it is neither the allocation nor the wire size: it is specifically handing a
large value *to* the host, and the asymmetry with returning it is the clue.

## Where the copies come from

For one `MontyObject::String` argument of N bytes, live bytes at the moment the
frame is encoded:

1. **N** — the Python `str` in the interpreter heap, still live (the suspended
   frame references it).
2. **N** — its owned `MontyObject` copy in `ReplFunctionCall.args`, built when
   the VM exit is converted (`crates/monty/src/repl.rs:1010`).
3. **N** — `suspension_event_function_call` **clones** the whole argument vector
   into the announcement (`crates/monty-proto/src/worker.rs`). The clone was
   deliberate: the suspension keeps its own args so a `Dump` of the suspended
   state (and its replay on `Load`) stays complete.
[...]

PR #699, opened 2026-08-09
Lend, don't clone, a suspension's payload when announcing it

samuelcolvin:
part of #698.

Announcing a suspension cloned the whole argument payload into the event, so a host-call argument was live four times over while the frame was encoded: the interpreter's own value, its converted `MontyObject`, the announcement's clone, and the encode buffer. Under `max_memory` that put the effective ceiling for a single argument at roughly a quarter of the budget — a 4 MiB string under a 10 MiB limit crossed the allocator's hard ceiling and the worker was killed mid-announcement. Returning the same value worked, because `complete_event` takes it by value.

`suspension_event_function_call` now `mem::take`s the args into the announcement, and `suspension_event_os_call` swaps the `OsFunctionCall` out for the unit `GetEnviron` variant; `Child::handle` moves the payload back into the stored suspension as soon as the event is sent, on the failed-send path too. The suspension still owns its payload — a `Dump` needs it — but only one copy of it exists at a time.

Peak is now 3N rather than 4N. The encode buffer is the remaining copy; removing it needs a streaming `BufMut` adapter, which `bytes` makes an unsafe trait impl, so it is left for #698 to decide.

Tests size the argument as a fraction of the limit so they cover the amplification rather than one absolute number, assert the return/pass symmetry that was broken, and dump a suspension immediately after announcing it so a payload left behind in the event is caught.

Claude-Session: https://claude.ai/code/session_01XS2wa7mvgYXA6Yquy1BNDu

---
## Summary by cubic
Lend, don't clone, suspension payloads when announcing function/OS calls to cut peak memory from 4N to 3N and prevent worker kills near `max_memory`. Restores the payload immediately so dumps/load stay correct and passing args now costs the same as returning values.

- **Bug Fixes**
  - Move payloads into events and reclaim them right after send (including failed sends) via `reclaim_suspension_payload`.
  - Function calls use `mem::take` for `args/kwargs`; OS calls swap the payload with unit `GetEnviron` while on the wire.
  - Add regression tests sizing args as a fraction of the limit and checking return/pass symmetry; update `limitations/resource_limits.md` to document the 3x headroom when crossing the host boundary.

<sup>Written for commit 5920d6b79a09306449e34d5c1956ec72822b6877. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/699?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty-runtime/tests/subprocess.rs
```diff
@@ -512,6 +584,84 @@ fn large_allocations_are_rejected_before_the_hard_limit() {
...
+    child.shutdown();
+}
+
+/// The asymmetry the amplification produced: a value small enough to *return*
+/// to the host was not necessarily small enough to *pass* to a host function,
+/// because only the announcement path cloned it. Both directions now cost the
+/// same, so one limit governs both.
+#[test]
+fn a_returnable_value_can_also_be_passed_to_a_host_function() {
+    const LIMIT: usize = 8 * 1024 * 1024;
```
rewitt94, line 624: Might be worth clarifying that this is the "max_memory", but the test does not actually go over the hard limit of "max_memory" + "headroom"

rewitt94: Also is it worth putting a test just below the threshold of "max_memory" + "headroom" so we ensure that limit does not reduce?

crates/monty-proto/src/worker.rs
```diff
@@ -849,30 +909,36 @@ fn oversize_suspension_error_message(event: &pb::ChildEvent) -> Option<String> {
...
+/// **moving** the call payload into it — see
+/// [`suspension_event_function_call`] for why, and
+/// [`Child::reclaim_suspension_payload`] for how it comes back. `GetEnviron` is
+/// the placeholder left behind: a unit variant, so the swap allocates nothing.
+fn suspension_event_os_call(call: &mut monty::ReplOsCall) -> pb::ChildEvent {
+    let function_call = mem::replace(&mut call.function_call, OsFunctionCall::GetEnviron);
     event(pb::child_event::Kind::OsCall(pb::OsCall {
         call_id: call.call_id,
-        call: Some(call.function_call.clone().into()),
+        call: Some(function_call.into()),
```
rewitt94, line 922: 👍

rewitt94 (approved):
LGMT! 

It's quite a sneaky memory expansion here - would be good to take this down to 2N if possible

Would be nice if we could somehow encode straight from the stored suspension, without having to pass ownership back and forth.

PR #702, opened 2026-08-10
Specialize list.append through PyTrait

KRRT7:
## Summary

Optimize exact one-positional-argument `.append(value)` calls without changing the bytecode encoding.

`CallAttr` checks the interned `append` name and argument count before heap access. Heap receivers then dispatch through `PyTrait::py_call_attr_one`: lists consume the argument directly, while every other type calls its normal `py_call_attr` implementation through the same heap read.

`ListAppend` remains comprehension-only, so existing dumps and the full `u8` comprehension-depth range keep their previous meaning. Keyword arguments, unpacking, and other arities retain normal attribute dispatch.

The receiver drop guard is created before its heap reader. A temporary receiver such as `[].append(1)` therefore releases the reader before decrementing the final reference.

## Performance

Local Criterion results against the rebased `main` baseline:

- `list_append_int__monty`: 6.85% faster, with a 95% interval of 5.44% to 7.93%
- `list_append_str__monty`: 5.50% faster, with a 95% interval of 3.38% to 6.98%

## Tests

Coverage includes ordinary list mutation, the `None` return value, heap-reference arguments, self-cycles, temporary receivers, user-defined `append` methods, and `collections.deque.append` fallback.

Validation:

- `cargo test -p monty`
- `cargo test -p monty --test inputs input_list_append --features memory-model-checks`
- `cargo run -p monty-datatest --features memory-model-checks -- list__ops`
- `cargo run -p monty-datatest --features memory-model-checks -- collections__deque`
- `cargo clippy -p monty --lib -- -D warnings -A clippy::unnecessary-box-returns`
- `cargo fmt --all -- --check`
- `git diff --check`

optimization found and verified by the [WarpForge](https://warpforge.co) agent

### DIFF
--- a/crates/monty/src/bytecode/op.rs
+++ b/crates/monty/src/bytecode/op.rs
@@ -32,10 +32,6 @@ pub const FORMAT_VALUE_STATIC_SPEC: u8 = 0x08;
 /// the opcode pops two comparison operands instead of one test value.
 pub const ASSERT_CMP_FLAG: u8 = 0x10;
 
-/// `ListAppend` operand selecting method-call semantics rather than a
-/// comprehension iterator depth.
-pub(crate) const LIST_APPEND_METHOD: u8 = u8::MAX;
-
 /// Encodes an optional fused comparison into the `Assert`/`AssertFailed`
 /// flags operand: `ASSERT_CMP_FLAG | as_operand` when present, `0` when not.
 pub fn assert_flags(cmp_op: Option<CmpOperator>) -> u8 {
@@ -266,9 +262,11 @@ pub enum Opcode {
     DictMerge = 71,
 
     // === Comprehension Building ===
-    /// Append TOS to a list. Operand: iterator depth or [`LIST_APPEND_METHOD`].
-    /// Comprehension stack: [..., list, iter1, ..., iterN, value] -> [..., list, iter1, ..., iterN].
-    /// Method stack: [..., receiver, value] -> [..., None].
+    /// Append TOS to list for comprehension. Operand: u8 depth (number of iterators).
+    ///
+    /// Stack: [..., list, iter1, ..., iterN, value] -> [..., list, iter1, ..., iterN]
+    /// Pops value (TOS), appends to list at stack position (len - 2 - depth).
+    /// Depth equals the number of nested iterators (generators) in the comprehension.
     ListAppend = 72,
     /// Add TOS to set for comprehension. Operand: u8 depth (number of iterators).
     ///
--- a/crates/monty/src/bytecode/op.rs
+++ b/crates/monty/src/bytecode/op.rs
@@ -32,6 +32,10 @@ pub const FORMAT_VALUE_STATIC_SPEC: u8 = 0x08;
 /// the opcode pops two comparison operands instead of one test value.
 pub const ASSERT_CMP_FLAG: u8 = 0x10;
 
+/// `ListAppend` operand selecting method-call semantics rather than a
+/// comprehension iterator depth.
+pub(crate) const LIST_APPEND_METHOD: u8 = u8::MAX;
+
 /// Encodes an optional fused comparison into the `Assert`/`AssertFailed`
 /// flags operand: `ASSERT_CMP_FLAG | as_operand` when present, `0` when not.
 pub fn assert_flags(cmp_op: Option<CmpOperator>) -> u8 {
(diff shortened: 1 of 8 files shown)


crates/monty/src/bytecode/vm/call.rs
```diff
@@ -203,6 +203,23 @@ impl VM<'_> {
...
+    /// Mutates exact lists directly, falling back to normal attribute lookup
+    /// so user-defined `append` methods retain `CallAttr` semantics.
+    pub(super) fn exec_call_list_append(&mut self) -> Result<CallResult, RunError> {
+        let item = self.pop();
+        let receiver = self.pop();
+        if let Value::Ref(list_id) = &receiver
+            && let HeapReadOutput::List(mut list) = self.heap.read(*list_id)
+        {
+            let this = self;
+            defer_drop!(receiver, this);
```
rewitt94: This fails if the the receiver has no other references -
`defer_drop` attempts to drop at the end of the if clause but the heap read was declared first and still live. Now freeing an entry with an open reader hits an `assert!` and panics.

I can reproduce with 
```
[].append(1)
```

crates/monty/src/bytecode/vm/call.rs
```diff
@@ -203,6 +203,23 @@ impl VM<'_> {
...
+            list.append(this, item);
+            Ok(CallResult::Value(Value::None))
+        } else {
+            self.call_attr(receiver, StaticStrings::Append.into(), ArgValues::One(item))
```
rewitt94: We get a small performance gain `<list>.append` - but have a performance cost on `<other_type>.append` as we do a heap read and discard it. I'm not convinced we need this.

crates/monty/src/bytecode/vm/mod.rs
```diff
@@ -1253,10 +1253,15 @@ impl<'h> VM<'h> {
                     let depth = cached_frame.fetch_u8() as usize;
                     try_catch_sync!(self, cached_frame, self.set_extend(depth));
                 }
-                // Comprehension Building - append/add/set items during iteration
+                // Collection mutation
                 Opcode::ListAppend => {
```
rewitt94, line 1324: We're changing this `OpCode` to do two quite different things, the stack shape is quite different here. Not the end of the world, but again I don't know if we need this.

crates/monty/src/bytecode/vm/call.rs
```diff
@@ -203,6 +203,23 @@ impl VM<'_> {
         self.call_attr(obj, name_id, args)
     }
 
+    /// Mutates exact lists directly, falling back to normal attribute lookup
+    /// so user-defined `append` methods retain `CallAttr` semantics.
+    pub(super) fn exec_call_list_append(&mut self) -> Result<CallResult, RunError> {
+        let item = self.pop();
+        let receiver = self.pop();
+        if let Value::Ref(list_id) = &receiver
+            && let HeapReadOutput::List(mut list) = self.heap.read(*list_id)
```
rewitt94: One of the main objectives of `PyTrait` is move all the type specific logic out of the VM layer and dispatch to types. This seems to be a step in the other direction.

crates/monty/test_cases/list__ops.py
```diff
@@ -108,6 +108,15 @@
...
 assert len(lst) == 3
 
+
+class AppendMethod:
+    def append(self, value):
+        return value + 1
+
+
+# Specialized append syntax retains dispatch for non-list receivers.
+assert AppendMethod().append(41) == 42
```
rewitt94, line 123: We should have had a couple cases where it was a list too - or at least checked existing coverage.

rewitt94 (commented):
Thanks for the contribution!!
A few issues with this PR, and not sure if it's the direction we want to go though.

KRRT7:
converted to draft whilst I rework based on feedback

crates/monty/src/bytecode/op.rs
```diff
@@ -32,6 +32,10 @@ pub const FORMAT_VALUE_STATIC_SPEC: u8 = 0x08;
 /// the opcode pops two comparison operands instead of one test value.
 pub const ASSERT_CMP_FLAG: u8 = 0x10;
 
+/// `ListAppend` operand selecting method-call semantics rather than a
+/// comprehension iterator depth.
+pub(crate) const LIST_APPEND_METHOD: u8 = u8::MAX;
```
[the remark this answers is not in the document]
KRRT7: _Auto response from GPT-5 Codex running `fix-pr-comments`:_ This is valid. The replacement implementation removes `LIST_APPEND_METHOD` entirely and leaves `ListAppend` comprehension-only, so existing dumps and depth 255 retain their original meaning. Exact one-argument `.append(...)` calls now use the existing `CallAttr` encoding, gated by cheap name/arity checks and dispatched through `PyTrait`. Regression tests and dump-format tests pass locally; these changes are currently uncommitted and will appear on the PR once the branch is pushed.

davidhewitt:
Thanks for the PR. I'm going to close this one; I don't think it's a good idea to specialize the VM around a single function. Very keen to find general systematic wins, this one seems too targeted to hit a function exercised by the existing benchmark suite.

PR #704, opened 2026-08-10
Avoid stack overflow in nested asyncio task chains

### LINKED ISSUE #649: CREATED BY CLAUDE: stack overflow (SIGABRT) from unbounded `handle_exception` recursion on deep async task chains
Sandboxed code can abort the monty process with a stack overflow (`SIGABRT`) using ~200 bytes of source. A deep chain of `asyncio.gather` calls where the innermost coroutine raises, and no frame in the chain has an exception handler, drives unbounded self-recursion in `VM::handle_exception`.

This is **not a regression** — it reproduces identically on the merge base (`69f8a613`). Filing it because it was found while stress-testing control flow for sandbox-escape/resource-limit issues, and it is not currently documented in `limitations/asyncio.md`.

## Reproducer

```python
import asyncio


async def leaf():
    raise ValueError('boom')


async def chain(n):
    if n == 0:
        return await asyncio.gather(leaf())
    return await asyncio.gather(chain(n - 1))


try:
    await asyncio.gather(chain(4000))
except ValueError as e:
    print('caught', e)
```

**Expected:** `caught boom`, or a catchable `RecursionError`.

**Actual:**

```
thread 'main' (21931285) has overflowed its stack
fatal runtime error: stack overflow, aborting
exit=134
```

The crash threshold is roughly depth 3000–4000 on current `HEAD` and ~5000 on the merge base — a frame-size difference, not a behavioural one.

## Root cause

`crates/monty/src/bytecode/vm/exceptions.rs:330`, in `handle_exception`:

```rust
Err(waiter_error) => {
    // Switched to waiter - handle error in waiter's context
    return self.handle_exception(waiter_error);
}
```

When a spawned task fails and no frame in that task has a matching entry in its exception table, `handle_task_failure` switches to the task's waiter and returns `Err(waiter_error)`, which re-enters `handle_exception`. Each link in the gather chain adds one Rust stack frame, and `handle_exception` is a large frame (it holds a `DropGuard` and the handler-search loop).

Crucially, this cascade is **not charged against `VM::recursion_depth`**, so no `RecursionError` is raised before the native stack is exhausted. An `lldb` backtrace shows a single function self-recursing thousands of frames deep at an identical return offset, consistent with this being the only self-recursive call on the path.

## Controls isolating the mechanism

Each of these is the *same* chain shape, and each survives depth **100,000**, which narrows the fault to the unhandled-exception cascade specifically:

| Variant | Depth 100,000 |
| --- | --- |
| Chain returns normally (no exception raised) | ok |
| Chain raises, but each level wraps the `await` in `try/finally` | ok — `caught boom` |
| Chain raises with no handler in any frame | **crashes at ~3500** |

The `try/finally` variant survives because a handler *is* found in the frame's exception table, so `handle_exception` returns `None` instead of recursing to the waiter.

## Impact

Not a sandbox escape — nothing is read or written out of bounds, and the guard page does its job. The severity depends on how the host embeds monty:

- **`pydantic_monty` / `monty-pool` (the only Python execution surface): contained.** Verified end to end — the parent raises `MontyCrashedError: monty worker crashed while waiting for a reply (signal: 6 (SIGABRT))`, discards the child, replaces it, and the pool remains usable (a subsequent `1 + 1` returns `2` on a fresh checkout). This is the documented crash-detection contract working as designed.
- **In-process Rust API (`MontyRun`) and the `monty` CLI: aborts the host process.** No opportunity to catch or clean up.

So for pool-based hosts this is a recoverable single-worker DoS (untrusted code can cheaply force worker churn); for in-process embedders it takes down the host. This is consistent with the note in `CLAUDE.md` that a monty process can never be made fully crash-proof against stack overflow, which is exactly why subprocess isolation exists — but the trigger here is cheap, deterministic, and reachable from ordinary sandboxed source.

## Suggested fix

The recursive call is a genuine tail call, so it can be rewritten mechanically as a loop: wrap the body of `handle_exception` so that the waiter-switch case reassigns `error = waiter_error` and continues the outer loop instead of recursing. That makes the cascade run in constant stack regardless of chain depth.

Alternatively (or additionally), charging the waiter cascade against `recursion_depth` would convert the abort into a catchable `RecursionError`, matching what plain recursion, `with`-block recursion, and `try/finally` recursion already do.

Either way, `limitations/asyncio.md` is worth a note: it currently documents `gather` semantics but says nothing about nesting depth limits or crash behaviour on deep task chains.

## Environment

- `pydantic/monty` version `0.0.19`, commit `f183c9d004859b334e45dd4150adaefb61e5508e`; also reproduced on merge base `69f8a613`.
- macOS (Darwin 25.5.0), aarch64.
- **Debug build.** `MAX_NESTING_DEPTH` is 30 in debug vs 200 in release, and release frames are smaller, so the exact crash threshold will differ in release — the recursion itself is unbounded in both.

---

jay3cx:
The waiter-error path re-entered `VM::handle_exception` recursively, adding one native Rust stack frame for each nested `asyncio.gather` waiter.

This changes waiter error propagation to an iterative loop while preserving the existing exception handling and ownership behavior.

A regression test covers a 4,000-level nested `asyncio.gather` chain and verifies that the original `ValueError` remains catchable.

## Tests

- `cargo test -p monty --test asyncio`
- `cargo test -p monty`

Fixes #649

---
## Summary by cubic
Prevents stack overflow when exceptions propagate through deep `asyncio.gather` waiter chains by switching from recursive to iterative handling. Preserves exception behavior; adds a regression test.

- **Bug Fixes**
  - Made `handle_exception_with_value` iterative via a loop over `handle_exception_step`.
  - Added `ExceptionHandlingResult` with `Caught`, `Unhandled`, and `PropagateToWaiter` to control propagation and handle waiter handoff without growing the stack.
  - Added a regression test for a 4,000-level nested `asyncio.gather`; verifies the original `ValueError` is catchable.

<sup>Written for commit 4f9202499a566a04a58fdb6519e0a942bf26bab3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/704?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/bytecode/vm/exceptions.rs
+++ b/crates/monty/src/bytecode/vm/exceptions.rs
@@ -14,6 +14,14 @@ use crate::{
     value::Value,
 };
 
+/// Result of handling an exception until execution can resume, it escapes, or
+/// it needs to be propagated in a waiting task.
+enum ExceptionHandlingResult {
+    Caught,
+    Unhandled(RunError),
+    PropagateToWaiter(RunError),
+}
+
 impl VM<'_> {
     /// Returns the current function name, or `<module>` outside a function.
     /// The empty-stack fallback keeps traceback generation total after async
@@ -246,8 +254,19 @@ impl VM<'_> {
     pub(super) fn handle_exception_with_value(
         &mut self,
         mut error: RunError,
-        raised: Option<Value>,
+        mut raised: Option<Value>,
     ) -> Option<RunError> {
+        loop {
+            match self.handle_exception_step(error, raised.take()) {
+                ExceptionHandlingResult::Caught => return None,
+                ExceptionHandlingResult::Unhandled(error) => return Some(error),
+                ExceptionHandlingResult::PropagateToWaiter(waiter_error) => error = waiter_error,
+            }
+        }
+    }
+
+    /// Handles one propagation step, yielding when the error moves to a waiter.
+    fn handle_exception_step(&mut self, mut error: RunError, raised: Option<Value>) -> ExceptionHandlingResult {
         // Ensure exception has initial frame info
         error = self.attach_frame_to_error(error);
 
@@ -257,7 +276,7 @@ impl VM<'_> {
             if let Some(raised) = raised {
                 raised.drop_with(self);
             }
-            return Some(self.unwind_for_traceback(error));
+            return ExceptionHandlingResult::Unhandled(self.unwind_for_traceback(error));
         }
 
         let exc_value = if let Some(raised) = raised {
@@ -320,7 +339,7 @@ impl VM<'_> {
                 // Jump to handler
                 this.current_frame_mut().ip = handler_offset;
 
-                return None; // Continue execution at handler
+                return ExceptionHandlingResult::Caught;
             }
 
             // No handler in this frame - pop frame and try outer
@@ -333,19 +352,16 @@ impl VM<'_> {
 
                 // For spawned tasks, fail the task instead of propagating
                 if is_spawned {
-                    match self.handle_task_failure(error) {
+                    return match self.handle_task_failure(error) {
                         Ok(()) => {
                             // Switched to next task - continue execution
-                            return None;
+                            ExceptionHandlingResult::Caught
                         }
-                        Err(waiter_error) => {
-                            // Switched to waiter - handle error in waiter's context
-                            return self.handle_exception(waiter_error);
-                        }
-                    }
+                        Err(waiter_error) => ExceptionHandlingResult::PropagateToWaiter(waiter_error),
+                    };
                 }
 
-                return Some(error);
+                return ExceptionHandlingResult::Unhandled(error);
             }
 
             // Get the caller's call-site offset before popping frame.
@@ -356,7 +372,7 @@ impl VM<'_> {
             if this.pop_frame() {
                 // The frame indicated evaluation should stop - e.g. inside `evaluate_function` - return the error
                 // now to stop unwinding.
-                return Some(error);
+                return ExceptionHandlingResult::Unhandled(error);
             }
 
             // Add caller frame info to traceback (if we have a call site).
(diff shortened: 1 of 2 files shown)


crates/monty/src/bytecode/vm/exceptions.rs
```diff
@@ -14,6 +14,14 @@ use crate::{
     value::Value,
 };
 
+/// Result of handling an exception until execution can resume, it escapes, or
+/// it needs to be propagated in a waiting task.
+enum ExceptionHandlingResult {
+    Caught,
```
rewitt94, line 21: These could have some concise doc strings maybe

crates/monty/tests/asyncio.rs
```diff
@@ -968,3 +968,34 @@ await main()
...
+except ValueError:
+    caught = True
+caught
+";
+
+    let runner = MontyRun::new(code.to_owned(), "test.py", vec![], CompileOptions::default()).unwrap();
+
+    let result = runner
+        .run_no_limits(vec![])
+        .expect("the original exception should be caught");
```
rewitt94: > the original exception should be caught

Maybe `should complete` would be better

rewitt94 (approved):
Thanks for contributing! This looks good to me!

crates/monty/src/bytecode/vm/exceptions.rs
jay3cx: Added concise documentation for each outcome variant in 4f920249.

crates/monty/tests/asyncio.rs
jay3cx: Updated the expectation message to `should complete` in 4f920249.

jay3cx (commented):
Done

jay3cx:
@rewitt94  Review comments are in 4f92024. Mind taking another look when you have a minute?

rewitt94:
Thanks @jay3cx! I will forward this to @davidhewitt

Issue #712, opened 2026-08-12
External/OS call in a sort key raises past `try`/`except` inside the key function


samuelcolvin:
(written by claude - hackmonty round 2 submission)

An external function or OS call inside a `sort()` / `sorted()` / `min()` / `max()` key function raises `NotImplementedError`, but the exception is constructed by the *caller* (`unsupported_frame_exit`, `crates/monty/src/bytecode/vm/call.rs:450`) after the callback's frames have already been popped. So it cannot be caught inside the key function, unlike any other Python exception.

```python
def key_fn(x):
    try:
        open('/data/sample.txt')
    except Exception as e:
        print('caught:', e)     # never runs
    return 0

[1, 2].sort(key=key_fn)
# NotImplementedError: sorted() key argument: OS function 'open' is not yet supported in this context
```

Sandboxed code has no way to defend against this — the failure always aborts the whole sort, and the `try`/`except` is silently inert.

Minor, same message: it says `sorted()` even for `list.sort()`, `min()` and `max()`.

This is also the trigger for #711.

PR #722, opened 2026-08-12
Arm `pending_os_effect` only at the host boundary

### LINKED ISSUE #711: Stale `pending_os_effect` after a rejected sort-key callback corrupts unrelated results
(written by claude - hackmonty round 2 submission)

When a `sort()`/`sorted()`/`min()`/`max()` key callback triggers a buffered file read, `run()` installs `pending_os_effect` before returning `FrameExit::OsCall`. The synchronous-context path then rejects that suspension in `unsupported_frame_exit` (`crates/monty/src/bytecode/vm/call.rs:450`), which drops the `FrameExit` but never clears `self.pending_os_effect` or the file's `pending_read`.

The next `resume()` (`crates/monty/src/bytecode/vm/mod.rs:1898`) diverts **any** result — an unrelated file read, an external function return, an OS call — into that stale file buffer.

## Repro

With `/data` mounted read-only:

```python
from pathlib import Path

f = open('/data/sample.txt')
try:
    [1].sort(key=lambda x: f.read(5))
except NotImplementedError:
    pass

Path('/data/config.json').read_text()   # -> '{\n  "'  — silently truncated to 5 chars
```

Type confusion — `read_text()` returns a `list`:

```python
f = open('/data/sample.txt')
try:
    [1].sort(key=lambda x: f.readlines())
except NotImplementedError:
    pass

r = Path('/data/config.json').read_text()
type(r).__name__   # 'list'
```

All confirmed live:

| after poisoning | result |
|---|---|
| `Path(...).read_text()` | truncated to the stale spec's length |
| `readlines()` spec, then `read_text()` | returns a `list`, not a `str` |
| external fn returning `'EXPLO_CONTENT_INJECTED'` | sandbox receives `'EXPLO'` |
| `Path(...).read_bytes()` into a text-mode buffer | `Internal error in monty: populate_buffer_meta: buffer type does not match file mode` |
| `Path(...).exists()` (bool result) | `Internal error in monty: apply_buffer_store: OS result must be a string or bytes value` |

## Scope

Confined to a single feed: `Drop for VM` clears the field, and I confirmed the state survives neither a feed boundary nor a checkout boundary (same worker pid reused from the pool, victim read came back correct). Values stay dynamically dispatched, so there is no Rust-level type/stride confusion — every operation on the confused `list` fails cleanly (`AttributeError`, `TypeError`, and `r[99999]` gives `IndexError`), and the two internal errors above are guards returning `Err`, not panics.

So: not a sandbox escape and not cross-user. It is silent corruption of values crossing the host boundary — a host that trusts the declared return type of its own external function, or the `str` from `read_text()`, gets a wrong type or a truncated value with no error anywhere.

## Fix

Clear `pending_os_effect` (and `dec_ref` the pinned file / clear its `pending_read`) in `unsupported_frame_exit` and `unsupported_call_result` — the same cleanup `Drop for VM` already performs.

rewitt94:
A file read inside a synchronously-evaluated callback (a `sorted()` key, `map`, an implicit dunder) armed `VM::pending_os_effect` as the `FrameExit::OsCall` was built, then had that exit rejected — leaving the effect to hijack the *next* OS result. Fixed by arming only once the call reaches the host.

Three notes, one root cause:

- **Not a sandbox escape** — silent corruption across the host boundary: `read_text()` truncated, or returning a `list`. One feed only.
- **Half-solved already** — `CallResult::OsCallWithEffect` carries its effect in the value; this applies that shape to `FrameExit`.
- **A third abandon site** — `frame_exit_to_object` leaked the same pin; all three now release through the owner's `drop_with`.

**Tests:** three in `os_tests.rs`, written failing first.

Fixes https://github.com/pydantic/monty/issues/711.

---
## Summary by cubic
Arm `VM::pending_os_effect` only when an OS call reaches the host, and release any stranded effect on every exit to prevent result corruption and file-pin leaks. Previously we armed on `FrameExit::OsCall` construction; rejected exits or `ExtFunctionResult::Future` replies left stale effects that mutated the next suspension’s result and hit a debug-only assert.

- `FrameExit::OsCall` carries `effect: Option<PendingOsEffect>`; `convert_frame_exit` now arms it at the host boundary and first clears any already-armed, stranded effect (covers non-OS suspensions too).
- Add `release_pending_effect`; call it from `CallResult::drop_with`, `FrameExit::drop_with`, `convert_frame_exit` (pre-arm clear), and `Drop for VM`.
- Restructure `frame_exit_to_object`: early-return on `Return`, build errors from borrows, then one `drop_with` to release all fields safely.
- Tests cover sync-context rejections and future-answered OS calls (no stranded effect, no reshaped results); document the sync-context I/O rejection in `limitations/open.md`.

<sup>Written for commit 60c6f39c27cac0128ffd8169dd84e9ac8bb8a939. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/722?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/run.rs
+++ b/crates/monty/src/run.rs
@@ -545,36 +545,34 @@ impl Executor {
 /// Used by non-iterative execution paths where suspendable outcomes (external calls,
 /// name lookups) are not supported and should produce errors.
 pub(crate) fn frame_exit_to_object(frame_exit_result: RunResult<FrameExit>, vm: &mut VM<'_>) -> RunResult<MontyObject> {
-    match frame_exit_result? {
-        FrameExit::Return(return_value) => Ok(MontyObject::new(return_value, vm)),
-        // Suspensions this path cannot service. Built from a borrow so one
-        // `drop_with` releases whatever the exit owns, fields added later too.
-        exit => {
-            let error = match &exit {
-                FrameExit::Return(_) => unreachable!("returns are handled above"),
-                FrameExit::ExternalCall { function_name, .. } => {
-                    let function_name = function_name.as_str(vm.interns);
-                    ExcType::not_implemented(format!(
-                        "External function '{function_name}' not implemented with standard execution"
-                    ))
-                }
-                FrameExit::OsCall { function_call, .. } => ExcType::not_implemented(format!(
-                    "OS function '{}' not implemented with standard execution",
-                    function_call.name()
-                )),
-                FrameExit::MethodCall { method_name, .. } => {
-                    let name = method_name.as_str(vm.interns);
-                    ExcType::not_implemented(format!("Method call '{name}' not implemented with standard execution"))
-                }
-                FrameExit::ResolveFutures(_) => {
-                    ExcType::not_implemented("async futures not supported by standard execution.")
-                }
-                FrameExit::NameLookup { name_id, .. } => ExcType::name_error(vm.interns.get_str(*name_id)),
-            };
-            exit.drop_with(vm);
-            Err(error.into())
+    // Suspensions this path cannot service. The error is built from a borrow
+    // so one `drop_with` releases whatever the exit owns, fields added later
+    // included.
+    let exit = match frame_exit_result? {
+        FrameExit::Return(return_value) => return Ok(MontyObject::new(return_value, vm)),
+        exit => exit,
+    };
+    let error = match &exit {
+        FrameExit::Return(_) => unreachable!("returns are handled above"),
+        FrameExit::ExternalCall { function_name, .. } => {
+            let function_name = function_name.as_str(vm.interns);
+            ExcType::not_implemented(format!(
+                "External function '{function_name}' not implemented with standard execution"
+            ))
         }
-    }
+        FrameExit::OsCall { function_call, .. } => ExcType::not_implemented(format!(
+            "OS function '{}' not implemented with standard execution",
+            function_call.name()
+        )),
+        FrameExit::MethodCall { method_name, .. } => {
+            let name = method_name.as_str(vm.interns);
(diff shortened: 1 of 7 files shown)


crates/monty/src/run.rs
```diff
@@ -547,38 +547,32 @@ impl Executor {
 pub(crate) fn frame_exit_to_object(frame_exit_result: RunResult<FrameExit>, vm: &mut VM<'_>) -> RunResult<MontyObject> {
     match frame_exit_result? {
```
samuelcolvin: i think we can write this slightly more cleanly as 

```rust
let error = match frame_exit_result? {
    FrameExit::Return(return_value) => return Ok(MontyObject::new(return_value, vm)),
    ...
};
exit.drop_with(vm);
Err(error.into())
```

samuelcolvin (commented):
overall this looks good.

Have you run `/review-security` on it? If not, you should. 

Also, does this fix #712?

samuelcolvin:
#725 was my AI trying to fix this, I would prefer to use your PR, but probably worth asking your AI if there's anything useful to pull from #725.

rewitt94:
Thanks! 👀

rewitt94:
The `/security-review` found an issue that we should always pre-emptively clear an existing `pending_os_effect` incase there was a previous call where the host responded with a future and it wasn't actioned (solved in commit 2/3).

#712 has a similar trigger scenario but is slightly different I'll raise a PR for that too

PR #750, opened 2026-08-14
Drop map values on error paths

smigolsmigol:
## Summary

`map()` accumulated heap-owned values in native vectors without contextual cleanup. If a later iterator or callback returned an error, those values could remain unreachable instead of being decremented.

Guard the partial output and per-call argument vectors so error and shortest-iterator paths drop owned values exactly once. The two-iterator fast path also guards its first argument while advancing the second iterator. Successful `map()` results are unchanged.

## Tests

- `cargo test -p monty --test resource_limits --features ref-count-return,memory-model-checks map_`
- `cargo +nightly fmt --all -- --check`
- `make lint-rs`

---
## Summary by cubic
Prevents `map()` from leaking heap-owned values when a callback or any iterator errors, or when the shortest iterator ends. Previously, partially accumulated outputs and per-call arguments could become unreachable without decref; now they are dropped exactly once. Successful results are unchanged.

- Guard the output vector and per-call argument vector with `defer_drop_mut!`; move them with `mem::take` only on success.
- In the two-iterator path, wrap the first argument in `DropGuard` while advancing the second iterator to ensure cleanup if it errors or exhausts.
- Add regressions for callback failure, iterator failure, and shortest-iterator exhaustion, and cover a successful three-iterator call before exhaustion to verify argument transfer and result correctness.

<sup>Written for commit 5dba24d8c38b6ff6325febf0023b7ca32f53bbc7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/750?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/builtins/map.rs
+++ b/crates/monty/src/builtins/map.rs
@@ -7,7 +7,7 @@ use crate::{
     bytecode::VM,
     defer_drop, defer_drop_mut,
     exception_private::{ExcType, ExcTypeExt, RunResult},
-    heap::{DropWithContext, HeapData},
+    heap::{DropGuard, DropWithContext, HeapData},
     types::{List, iter::checked_preallocation_hint},
     value::Value,
 };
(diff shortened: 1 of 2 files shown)


davidhewitt (approved):
Thanks, will merge this however I think in the "near" future we'll make `map` a proper iterator.

PR #762, opened 2026-08-19
remove `CachedFrame` concept

davidhewitt:
vibe-coded experiment to see what happens if we don't need to "sync" the IP in case of re-entrancy...

---
## Summary by cubic
Removes `CachedFrame` and runs the VM from a single current frame with suspended callers. This removes IP sync on re-entrancy, routes exceptions/awaits to the active opcode frame, and lets REPL `call_function` run without a host frame.

- The run loop advances the current frame IP directly; no cache reload or IP sync before host/external calls.
- `CallFrame` caches a bytecode slice to remove a dependent load and speed opcode dispatch.
- `current_frame` plus `suspended_frames` replace frame caching; recursion depth ignores a host root, and `call_function` no longer creates one.
- Exception routing/traceback use the authoritative `instruction_ip` and the `current_frame`.
- Async save/restore serializes suspended callers and the current frame; task switching reconstructs both and parks the placeholder frame when no task is ready.
- Updates REPL/Executor/tests: `call_function` can run `asyncio.gather`; adds tests ensuring re-entrant comparison exceptions surface at the correct frame.

**Migration**
- Construct VMs with code: call `VM::new(globals, &module_code, …)`; call `run_module()` with no arguments.
- For host-initiated calls, pass a `Code` to `VM::new` (in tests use `Code::empty()`); thread a `&Code` through `HeapReader::with` closures.

<sup>Written for commit efef21da51965f565fa4867bff6cf8028904ea89. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/762?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


davidhewitt:
@codspeedbot why the regression?

crates/monty/src/repl.rs
```diff
@@ -286,63 +288,72 @@ impl MontyRepl {
...
-                let result = match eval_result {
-                    Ok(value) => Ok(MontyObject::new(value, vm)),
-                    Err(e) => {
-                        Err(e.into_python_exception(&self.interns, |fname| self.sources.get(fname).map(String::as_str)))
+            let result = match convert_args(args, vm) {
+                Ok(args) => {
+                    let (args, kwargs) = args.into_parts();
+                    debug_assert!(kwargs.is_empty(), "host function calls only have positional arguments");
+                    kwargs.drop_with(vm);
+                    let args_tuple = allocate_tuple(args.collect(), vm.heap);
```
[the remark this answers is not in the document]
davidhewitt: Not concerned about this path for now.

crates/monty/src/bytecode/vm/async_exec.rs
```diff
@@ -696,7 +709,7 @@ impl<'h> VM<'h> {
             return;
         }
 
-        let task_is_current = self.scheduler.current_task_id() == Some(task_id) && !self.frames.is_empty();
+        let task_is_current = self.scheduler.current_task_id() == Some(task_id);
```
rewitt94: We remove `&& !self.frames.is_empty()` - should it instead be `!self.current_frame.is_parked` ?

crates/monty/src/bytecode/vm/mod.rs
```diff
@@ -883,25 +865,26 @@ impl<'h> VM<'h> {
             // (the VM owned them, now the snapshot owns them)
             stack: mem::take(&mut self.stack),
             globals: mem::take(&mut self.globals),
-            frames: self.frames.iter().map(CallFrame::serialize).collect(),
+            frames: if self.current_frame.is_parked {
+                Vec::new()
```
rewitt94, line 873: `debug_assert!(self.suspended_frames.is_empty())` - should we assert this here as it's implied?

rewitt94 (approved):
🎉 Removing this cache like a good way to reduce the surface area for bugs! 
I guess the now reduced surface area is the management of the `is_parked` field.

Do we have enough test coverage for this (especially in the async section)

davidhewitt:
> I guess the now reduced surface area is the management of the `is_parked` field.
> 
> Do we have enough test coverage for this (especially in the async section)

Yes. I might try to remove that concept later but that requires hacking with async code and seems lower risk than the bytecode loop not supporting re-entrancy.

PR #764, opened 2026-08-19
Raise instead of aborting when the async scheduler stalls

### LINKED ISSUE #732: Panic: `resume_with_resolved_futures` assert — no pending calls and no ready tasks
_(Create by claude code)_

> Found in production panic telemetry from Hack Monty (Logfire project `hackmonty`, service `monty-server`, 24h to 2026-08-13 09:13Z, deployed rev `2680c32`): **3 panics**, and the most recent panic in the window (last seen 2026-08-13 08:57Z). Unrelated to the container-mutation panic family (#715/#716/#729). **Not reproduced locally yet** — see "What I tried" below.

## Summary

`VM::resume_with_resolved_futures` ends with an `assert!` declaring that if the current task cannot resume and no other task is ready, there must be at least one pending external call. Production hit that assert three times, spread across the window (01:46 and two later) rather than in a burst, which suggests a rare interleaving rather than one participant hammering it.

```
thread 'main' panicked at crates/monty/src/bytecode/vm/async_exec.rs:890:9:
resume_with_resolved_futures called but no pending calls and no ready tasks
```

The panic kills the worker mid-turn; `monty-server` catches the child death and replies `Crash`, so the server itself stays healthy and the session is lost.

## Site

`crates/monty/src/bytecode/vm/async_exec.rs:831` (`resume_with_resolved_futures`), post-loop:

```rust
let pending_call_ids = self.get_pending_call_ids();

assert!(                                                          // :890
    !pending_call_ids.is_empty(),
    "resume_with_resolved_futures called but no pending calls and no ready tasks"
);
```

Reaching it means: the scheduler had no current task (or the current task was `Blocked`), `next_ready_task()` returned `None`, and `pending_call_ids()` was empty — a state the assert declares impossible.

## Suspected mechanism

The `current_task_id() == None` route looks the most likely. `Scheduler::cancel_task` explicitly clears `current_task` and calls out this exact caller:

```rust
// If we're cancelling the current task, clear `current_task` so callers
// don't try to look up a task that's about to be dropped (e.g.
// `resume_with_resolved_futures` after `fail_for_call` tore down the
// gather containing the previously-current task).
```
(`crates/monty/src/bytecode/vm/scheduler.rs:397`)

So: an external future is settled with an error, `fail_for_call` tears down the gather that contained the *current* task, `cancel_task` clears `current_task` and retires the siblings' pending externals — and the surviving waiter is left in a state that is neither `Ready` nor current. The `if let Some(current_task_id)` block is then skipped, `next_ready_task()` finds nothing, and the pending set is empty.

Either a task can reach a state where it is neither ready, blocked, failed nor completed, or a pending call is being retired before the resume path observes it.

## What I tried (did not reproduce)

Driving `feed_start` / `FutureSnapshot.resume(...)` manually against a debug worker on `main` @ `d7a34fb0`, answering external calls with `{'future': ...}` and then settling selected `call_id`s with an exception:

- `gather(a(), b())` with two pending externals, failing one or both
- the same via child coroutines (`gather(c1(), c2())`, `gather(c1(), c2(), c3())`), failing each `call_id` in turn
- nested gathers (`gather(inner1(), inner2())`, each inner gathering two external-awaiting children), failing each of the four `call_id`s
- `try` / `except ValueError` around the gather and around the inner `await`

All produced the expected `MontyRuntimeError: ValueError: boom` (or the caught result) with no crash. A repro likely needs a specific interleaving — probably failing a call whose teardown cancels the currently-executing task while the remaining pending calls are retired in the same step.

## Suggested direction

Independent of finding the exact repro, this assert is a liveness invariant asserted on a host-driven path, so a malformed or unusual interleaving turns into a hard abort. Downgrading the impossible state to a `RunError` (e.g. a `RuntimeError` naming the inconsistent scheduler state) would turn a worker crash into a recoverable turn failure and would also make the state visible in telemetry rather than only in a panic message.

rewitt94:
Closes pydantic/monty#732.

Return an exception here as the invariant was reached by users. Meaning
**now**: end the turn as a RuntimeError
**before**: panic aborts the worker process so the pool has to replace it

VM::resume_with_resolved_futures` ended with an `assert!` that no ready tasks implies at least one pending external call; it now returns `RunError::Internal`, which the host receives as `RuntimeError: Internal error in monty: asyncio scheduler stalled: no ready tasks and no pending external calls`.

Note: 
Two tests covered previously untested adjacent code paths — an external orphaned by a cancelled task, resolved and failed alongside a live one. **However, a situation to reproduce the error is was not identified.** 

---
## Summary by cubic
Raise `RunError::Internal` instead of aborting the worker when the asyncio scheduler stalls. Before: an assert in `resume_with_resolved_futures` killed the worker and session; now only the current turn ends and the host sees `RuntimeError: Internal error in monty: asyncio scheduler stalled: no ready tasks and no pending external calls` (uncatchable by user code).

- `VM::resume_with_resolved_futures` returns `RunError::Internal` when no ready tasks and no pending externals; otherwise returns `ResolveFutures`. Worker and session stay alive.
- Adds order-independent tests for an orphaned external resolved or failed alongside a live one; keys call IDs by function name.
- Documents the stalled-turn `RuntimeError` in `limitations/asyncio.md`.
- Removes a dead Clippy `unnecessary_box_returns` expectation in `stable_heap::create_page` to fix a `-D warnings` build error.

<sup>Written for commit 3f8fc9c9b19e185cd438d0e0580876d984ab7d2e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/764?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/limitations/asyncio.md
+++ b/limitations/asyncio.md
@@ -54,6 +54,11 @@ every branch is blocked on an external call, hands the pending calls to the
 host, and resumes when the host returns results. There is no preemption, no
 threads, and no in-sandbox scheduler.
 
+If the scheduler ends a resume with nothing ready to run and no external calls
+outstanding, the turn fails with `RuntimeError: Internal error in monty:
+asyncio scheduler stalled: no ready tasks and no pending external calls`, which
+`try`/`except` cannot catch.
+
 ### A failing `gather` cancels its siblings
 
 When one child of a `gather` raises, every sibling still running is cancelled where it is blocked and never resumes.
--- a/limitations/asyncio.md
+++ b/limitations/asyncio.md
@@ -54,11 +54,6 @@ every branch is blocked on an external call, hands the pending calls to the
 host, and resumes when the host returns results. There is no preemption, no
 threads, and no in-sandbox scheduler.
 
-If the scheduler ends a resume with nothing ready to run and no external calls
-outstanding, the turn fails with `RuntimeError: Internal error in monty:
-asyncio scheduler stalled: no ready tasks and no pending external calls`, which
-`try`/`except` cannot catch.
-
 ### A failing `gather` cancels its siblings
 
 When one child of a `gather` raises, every sibling still running is cancelled where it is blocked and never resumes.
--- a/crates/monty/src/bytecode/vm/async_exec.rs
+++ b/crates/monty/src/bytecode/vm/async_exec.rs
@@ -828,6 +828,10 @@ impl<'h> VM<'h> {
     /// 2. Attempt to resume the current task (or fail it if any future resolution caused it to fail)
     /// 3. Load a ready task if needed (current task still blocked)
     /// 4. If no task is ready, return `ResolveFutures` with remaining pending call IDs
+    ///
+    /// # Errors
+    /// Returns [`RunError::Internal`] if nothing is ready to run and nothing is
+    /// pending: unreachable by design, but ends the turn rather than the worker.
     pub fn resume_with_resolved_futures(&mut self, results: Vec<(u32, ExtFunctionResult)>) -> RunResult<FrameExit> {
         for (call_id, ext_result) in results {
             match ext_result {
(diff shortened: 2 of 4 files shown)


limitations/asyncio.md
```diff
@@ -54,6 +54,11 @@ every branch is blocked on an external call, hands the pending calls to the
 host, and resumes when the host returns results. There is no preemption, no
 threads, and no in-sandbox scheduler.
 
+If the scheduler ends a resume with nothing ready to run and no external calls
+outstanding, the turn fails with `RuntimeError: Internal error in monty:
+asyncio scheduler stalled: no ready tasks and no pending external calls`, which
+`try`/`except` cannot catch.
+
```
davidhewitt: Don't think this is relevant to document.

```suggestion
```

davidhewitt (approved):

PR #769, opened 2026-08-22
Fix class-body comprehension scope

### LINKED ISSUE #755: Class-body comprehensions resolve names from class scope instead of global scope
For context, The following bug report was generated by codex 🤖.

## Summary

Monty 0.0.21 produces different behavior from CPython for comprehensions defined in a class body. A name loaded inside the comprehension resolves to the class attribute value in Monty, but CPython resolves it from the surrounding global/module scope.

This is a semantic mismatch, not just an unsupported feature diagnostic.

## Reproducer

```python
x = 10


class A:
    x = 20
    ys = [x for _ in range(1)]


print(A.ys)
```

## Steps to reproduce

Install Monty and run the reproducer through the public `pydantic_monty` API:

```bash
python -m pip install 'pydantic-monty==0.0.21'
python repro.py
```

`repro.py`:

```python
import pydantic_monty

code = """
x = 10


class A:
    x = 20
    ys = [x for _ in range(1)]


print(A.ys)
"""

with pydantic_monty.Monty(request_timeout=3) as pool:
    with pool.checkout(script_name="agent.py", type_check=False) as session:
        session.feed_run(code, inputs={"prompt": "testing"})
```

## Actual behavior

Monty prints:

```text
[20]
```

## Expected behavior

CPython 3.13 prints:

```text
[10]
```

## Why this matters

In Python, a class body has special name-resolution behavior. Nested functions and comprehensions created inside the class body do not close over the class namespace. In this example, the `x` loaded inside the list comprehension should resolve to the module-level `x = 10`, not the class attribute `A.x = 20`.

The same mismatch also appears with generator expressions:

```python
x = 10


class A:
    x = 20
    ys = list(x for _ in range(1))


print(A.ys)
```

CPython prints `[10]`; Monty prints `[20]`.

## Environment

- Monty package: `pydantic-monty==0.0.21`
- Python: CPython 3.13

simonyang08:
Fixes #755

## Summary

- evaluate only the first iterable of a class-body comprehension in the class namespace
- resolve filters, later iterables, and result expressions through the comprehension's nested scope
- keep enclosing-function captures distinct from same-name class members, including a class assignment after the capture

## Root cause

Monty prepared comprehensions inline with the surrounding class-body preparer. Later comprehension reads therefore continued through the class namespace instead of skipping it as CPython does.

After the initial scope fix, a comprehension could allocate a free-variable slot for an enclosing local before a later same-name class assignment. That store then reused the capture slot and failed to create the class member. Class-body stores now allocate a class-local slot before free-variable lookup; explicit `nonlocal` stores continue to use their cell.

## Validation

- class-scope CPython/Monty differential test: passed, including the reported post-capture class assignment and a same-name comprehension target
- `cargo test -p monty --lib`: 45 passed
- full datatest: 1,183 passed; the only failure is the pre-existing CPython `sequence__search_bounds` mismatch
- `make format-rs`: passed
- both workspace Clippy invocations from `make lint-rs`: passed with `-D warnings`
- `scripts/check_imports.py`: passed from the Windows worktree
- `git diff --check`: passed

The latest head passes the repository's full GitHub Actions matrix, including lint, Rust/Python/JS tests, coverage, Miri, fuzz, browser, and both benchmark jobs. CodSpeed separately flags one simulated benchmark at -11.17%, while warning that the comparison used different runtime environments; its external performance check remains red for maintainer review.

## AI assistance

This change was prepared with OpenAI Codex assistance and reviewed and validated locally by the contributor.

### DIFF
--- a/crates/monty/src/prepare.rs
+++ b/crates/monty/src/prepare.rs
@@ -269,6 +269,8 @@ struct FunctionState {
     locals: NameMap,
     /// Names declared `global` in this function — resolve to module globals.
     global_names: AHashSet<StringId>,
+    /// Names declared `nonlocal` in this function — resolve to enclosing cells.
+    nonlocal_names: AHashSet<StringId>,
     /// Names bound in THIS scope (params + body-assigned, minus globals).
     /// A read of any of these resolves to `NameScope::Local`.
     assigned_names: AHashSet<StringId>,
@@ -628,6 +630,7 @@ impl<'i, 'g> Prepare<'i, 'g> {
             state: PrepareState::Function(Box::new(FunctionState {
                 locals,
                 global_names,
+                nonlocal_names: nonlocal_names.clone(),
                 assigned_names,
                 enclosing_locals,
                 free_var_map,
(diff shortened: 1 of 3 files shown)


simonyang08:
Confirmed and fixed in bb7a0de. The class-body store now allocates a class-local slot before free-variable lookup, while explicit `nonlocal` stores remain cells.

The reported case now matches CPython (`body == [10]`, class member `x == 5`). I also added coverage for a comprehension target sharing its name with a later class assignment. The focused differential test, 45 Rust library tests, formatting, both Clippy passes, and the import check pass locally.

PR #772, opened 2026-08-23
Let a failing `gather` leave its siblings running

rewitt94:
When one child of a `gather` raises, the other children keep running instead of being cancelled, as they do in CPython.

-------

The reference chain is task→gather→task which remains unchanged (child points up at the gather)
- tasks in a hash map on the scheduler
- gathers live on the heap

**Change**: `Task::gather_id` becomes `Task::awaiter: Opt>` - `gather_id` was a special-case field needing bespoke handling; as an `Awaiter`, coroutine children join the generic chain walk, so "already settled?" is decided in one place. 

**Enables**: cancelling siblings was the old way to stop late results reaching a dead gather; now a single settled-check discards them instead, so siblings keep running.

Fixes three panics reachable from sandboxed Python: dispatch against an empty frame stack, `fail_future` discarding a live sibling's frames, and freeing a coroutine under a live reader.

Divergences: `limitations/asyncio.md`.

---
## Summary by cubic
Lets a failing `asyncio.gather` leave its siblings running instead of cancelling them, matching CPython. Results or exceptions arriving after the gather settles are discarded, and exceptions in detached tasks stay silent.

- Replaces `Task::gather_id` with `Task::awaiter: Option<Awaiter>` so each spawned child owns the edge to its gather; the failed gather is retained until its last child finishes and releases it.
- Settles a gather in place without touching its children: delivery walks the awaiter chain, and a value or error landing on an already-settled gather, or on no waiter, is dropped.
- Switches safely for detached siblings: `park_current_context` saves a parked sibling's live frames, `discard_failed_task` removes failed tasks, and `yield_if_parked!` yields to the host when only parked frames remain.
- Has `Scheduler::fail_for_call` settle an enclosing still-running gather and return its awaiter, and drops a gather child whose own external call failed so it doesn't stay parked forever on a future that can never resolve, releasing its references.
- Errors with the real cause when a discarded task leaves nothing to run — no pending calls yields "asyncio scheduler stalled" rather than blaming the scheduler one round-trip later.
- Fixes three panics reachable from sandboxed Python: dispatch against an empty frame stack, `fail_future` discarding a live sibling's frames, and freeing a coroutine under a live reader.
- Adds tests for detached siblings, nested gathers, externals, and refcounts — the latter verified through a `test-hooks` live-task-count bridge — finalizing the scheduler before ref-count checks, and updates `limitations/asyncio.md` for the lazy start, siblings advancing only when something else suspends, and unreported unretrieved exceptions.

<sup>Written for commit bda89fbf245fb2c54d663e8ce8f82f4ae6cf2be0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/772?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/bytecode/vm/async_exec.rs
+++ b/crates/monty/src/bytecode/vm/async_exec.rs
@@ -18,7 +18,7 @@ use crate::{
         AwaitedGather, Awaiter, CallId, Coroutine, CoroutineState, ExternalFuture, ExternalFutureState, GatherFuture,
         GatherState, TaskId,
     },
-    bytecode::vm::scheduler::{Scheduler, SerializedTaskFrame, TaskState},
+    bytecode::vm::scheduler::{SerializedTaskFrame, TaskState},
     defer_drop, defer_drop_mut,
     exception_private::{ExcType, ExcTypeExt, RunError, RunResult, SimpleException},
     heap::{DropGuard, DropWithContext, HeapData, HeapId, HeapObjectRead, HeapRead, HeapReadOutput, HeapReader},
@@ -140,12 +140,11 @@ impl<'h> VM<'h> {
         let mut results_guard = DropGuard::new(results, this);
         let (results, this) = results_guard.as_parts_mut();
 
-        // Roll back already-committed siblings if a later child fails during
-        // this commit pass; otherwise spawned tasks or awaiters can outlive a
-        // gather that never reached `Awaited`.
+        // A later item failing during this commit pass leaves the ones already
+        // committed running, as CPython does. They keep pointing at this
+        // gather; `resolve_child` drops what they deliver to a `Failed` one.
         if let Err(err) = this.commit_gather_items(&gather, &mut pending_children, results) {
             gather.get_mut(this.heap).state = GatherState::Failed(err.clone());
-            drop_committed_children(pending_children, &mut this.scheduler, this.heap, &err);
             return Err(err);
         }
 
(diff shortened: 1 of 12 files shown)


crates/monty/src/bytecode/vm/async_exec.rs
```diff
@@ -1011,81 +1050,43 @@ impl<'h> HeapRead<'h, GatherFuture> {
...
-        let (waiter, pending_children, results) = {
-            let awaited = self
-                .get_mut(heap)
-                .as_awaited_mut()
-                .expect("fail called on non-Awaited gather");
+        // — it owns any `GatherSlot` inc_ref it carried. Already settled means
+        // a chain that ends here: an earlier failure cached its error and
+        // handed the waiter on.
+        let (waiter, results) = {
+            let awaited = self.get_mut(heap).as_awaited_mut()?;
```
rewitt94, line 1075: This now returns an option

crates/monty/src/bytecode/vm/async_exec.rs
```diff
@@ -370,28 +367,18 @@ impl<'h> VM<'h> {
...
-        // and the final state transition; it must run BEFORE we release any
-        // inc_refs the gather is holding (cancelling children, dropping the
-        // waiter's `Blocked` ref) — otherwise the gather can be freed while
-        // we're still about to write its cached state. The gather keys by
-        // item HeapId, so we pass the coroutine's id rather than the
-        // (kind-specific) task id.
-        let HeapReadOutput::GatherFuture(mut gather) = self.heap.read(gid) else {
-            panic!("task gather_id doesn't point to a GatherFuture")
-        };
-        let resolution = gather.resolve_child(self, coroutine_id, result);
```
rewitt94, line 384: `resolve_child` is now exclusively called via `deliver_awaiter_success`

davidhewitt (approved):

PR #784, opened 2026-08-27
implement generator expressions

### LINKED ISSUE #364: Generator expression is materialized to list, breaking iterator semantics (next(...) fails)
First of all, **thank you** for Monty! It unlocks a step change in capability for our Pydantic AI agents without the operational headaches of full sandboxes. 

---

Monty appears to convert generator expressions (`ast.GeneratorExp`) into lists during evaluation, which breaks Python iterator semantics.  

### Minimal reproduction
```python
from pydantic_monty import Monty

print(Monty("type(v for v in [1, 2, 3])").run())
print(Monty("next(v for v in [1, 2, 3])").run())
```

### Actual behavior
- `type(v for v in [1, 2, 3])` -> `<class 'list'>`
- `next(v for v in [1, 2, 3])` -> `MontyRuntimeError: TypeError: 'list' object is not an iterator`

### Expected behavior (CPython parity)
- `type(v for v in [1, 2, 3])` -> `<class 'generator'>`
- `next(v for v in [1, 2, 3])` -> `1`

### Environment
- `pydantic_monty` version: `0.0.14`
- Python: `3.13` (via `uv`)

---

Context for this low-priority request: We're seeing our agents often run into execution failures finding a value in a list via `next`. Our models can recover transparently by rewriting the code into `for` loops - but it adds latency and token cost. I realize adding support for generators is likely a non-trivial improvement - but probably falls within the core Python language feature set Monty would ideally support.

davidhewitt:
Fixes #364 

---
## Summary by cubic
Implements generator expressions as lazy, single-pass iterators instead of eager list comprehensions, matching CPython's scoping, closure capture, walrus binding, PEP 479, and re-entry errors. Fixes #364.

**Implementation**
- Adds a `Generator` heap payload that owns a suspended `<genexpr>` frame, created by `MakeGenerator` and resumed by `YieldValue`.
- Compiles the lazy body into a synthetic child scope; only the outermost iterable is consumed at creation.
- Generators are refcounted, participate in cycle collection, and survive REPL dump/load with tracebacks preserved across the generator boundary.

**Divergences**
- `send`, `throw`, `close`, and generator attributes are not implemented; async bodies are rejected at compile time.
- Advancement cannot suspend to the host, and the PEP 479 `RuntimeError` merges body and consumer frames.

<sup>Written for commit d65ca9bef5448a374b66486d53417b630bbf5176. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/784?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/src/bytecode/compiler.rs
```diff
@@ -3112,7 +3204,7 @@ impl<'a> Compiler<'a> {
 
         // Recurse or emit body.
         if index + 1 < generators.len() {
-            self.compile_comprehension_generators(generators, index + 1, body_fn)?;
+            self.compile_comprehension_generators(generators, index + 1, first_iterator_ready, body_fn)?;
```
rewitt94: `first_iterator_ready` could be an explicit false here as index > 0?

rewitt94 (commented):
Looks good to me -

There seem to be a couple out of date docs to update:
https://github.com/pydantic/monty/blob/3268fea2196e1cba085fa743e0210f0cca15468e/docs/python-subset.md?plain=1#L39

https://github.com/pydantic/monty/blob/3268fea2196e1cba085fa743e0210f0cca15468e/docs/index.md?plain=1#L68-L69

PR #805, opened 2026-09-03
Add a suspension limit to ResourceLimits

### LINKED ISSUE #736: Add a `max_suspensions` limit to `ResourceLimits`
(written by claude)

## Summary

`ResourceLimits` bounds memory, execution time and recursion depth, but nothing bounds how many times a session may **suspend**. A snippet that catches each refusal and immediately retries produces an unbounded stream of parent-side round trips while making no progress and producing no output:

```python
while True:
    try:
        open('/etc/passwd')
    except Exception:
        pass
```

Each iteration is a full round trip — a suspension frame to the parent, a span recorded, a resume frame back — and every one costs the parent retained memory (~4 KB per suspension span, measured; larger than the ~800 bytes a print record costs, see #734).

## Why the existing limits do not cover it

This is the part that makes it worth a limit of its own rather than tuning an existing one:

- **`max_memory`** — the sandbox allocates nothing; the growth is entirely parent-side.
- **`max_duration`** — measures bytecode execution, which is *paused while suspended*. A loop that spends nearly all its wall clock waiting on the parent barely accrues any, so this limit effectively never fires for this shape.
- **output limits** — the snippet prints nothing.

That leaves only whatever whole-turn deadline the host happens to impose. In `monty-pool` terms there is no such thing; it is entirely the embedder's problem, and an embedder that answers host calls quickly (which is the normal case — that is what a host *is*) gets the accumulation at full speed.

Measured on a host with a 6s turn deadline, the snippet above cost **+2127 MB of parent RSS from a single run**. With no deadline it is bounded by nothing at all.

## Proposal

Add a suspension budget to `ResourceLimits` — `max_suspensions`, counting the suspensions a session may have answered for it, in the same "the interpreter enforces it and its error quotes it" style as the existing limits.

Points worth deciding:

- **Per session or per turn?** Per session matches `max_duration`'s cumulative framing and is harder to sidestep by feeding repeatedly; per turn is easier to reason about for a host. Cumulative-per-session seems right, for the same reasons `max_duration` is.
- **What it counts.** All suspension kinds — external calls, name lookups and OS calls — since all three are parent round trips with a span each.
- **What happens on exhaustion.** An exception naming the limit, so it behaves like `MemoryError` / the time limit rather than silently truncating. Code that catches everything will catch this too, but the budget is spent by then, so the next suspension re-raises immediately and the loop cannot continue.

## Workaround in the meantime

A host can cap it itself by counting suspensions in its own resume loop and abandoning the checkout when the count is exceeded. That works, but it puts a security-relevant bound in every embedder rather than in the interpreter, and each one has to rediscover that neither `max_duration` nor the memory limit covers this.

## Related

#734 — the same "one retained parent-side record per sandbox event" shape, via prints rather than suspensions. Debouncing prints does not touch this case, and a suspension limit does not touch that one; they are siblings.

rewitt94:
Closes #736 by adding a suspension budget to `ResourceLimits` — `max_suspensions_per_run` (always on, default 10,000) and the opt-in cumulative `max_total_suspensions` — charged in `convert_frame_exit` before an over-budget suspension is announced, so a sandbox loop that repeatedly suspends and swallows the result can no longer grow the host's retained state without bound.

---
## Summary by cubic
Adds a suspension budget to `ResourceLimits` so sandboxed code that repeatedly suspends to the host can no longer grow host-retained state without bound (fixes #736). `max_suspensions_per_run` is always on (default 10,000) and `max_total_suspensions` is an opt-in cumulative cap across the session.

**Behavior**
- Each suspension is charged in `convert_frame_exit` before the over-budget one is announced, so the host never pays for a refused round trip.
- Exceeding either limit ends the run with an uncatchable `RuntimeError` naming the limit.
- The per-run counter resets at each feed; the total counter never resets.
- Both counters are serialized into dumps so a restored session resumes with its budget spent.

<sup>Written for commit 0b240a326e80610e6a48f40740552c00b415a878. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/805?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -959,8 +959,8 @@ where they are. Change one and you must change all of them:
 
 - **The importable stdlib module list** — `limitations/modules.md` (authoritative),
   `docs/python-subset.md`, `docs/index.md`, `README.md`.
-- **Default resource limits** (1000 recursion frames, 100 MB per-mount memory, 10 MiB
-  print collectors, 1s duration grace) — `limitations/resource_limits.md`,
+- **Default resource limits** (1000 recursion frames, 10,000 suspensions per run,
+  100 MB per-mount memory, 10 MiB print collectors, 1s duration grace) — `limitations/resource_limits.md`,
   `docs/resource-limits.md`, and the binding docstrings.
 - **Mount modes and their defaults** — `limitations/filesystem.md`, `docs/filesystem.md`,
   the `MountDir` docstrings in `_monty.pyi` and `crates/monty-js/ts/mount.ts`.
--- a/crates/monty-js/README.md
+++ b/crates/monty-js/README.md
@@ -256,6 +256,14 @@ const limited = await pool.checkout({
 interpreter itself: the worker is killed and the session fails with
 `MontyCrashedError` (`timedOut: true`).
 
+`maxSuspensionsPerRun` (default 10,000) bounds how many times one run may
+suspend to the host — an external function call, a name lookup or an `os`
+callback. Each is a round trip the host retains state for, and none of it is
+visible to `maxMemory` or `maxDurationSecs`, so a loop that suspends and
+swallows the result is bounded by nothing else. `maxTotalSuspensions` is the
+optional cumulative cap across the session. Exceeding either ends the run with
+an uncatchable `RuntimeError` naming the limit.
+
 `maxDurationSecs` limits cumulative _execution_ time: the sandbox clock runs
 only while the interpreter executes, never while suspended waiting on an
 external function or between feeds. Sessions with the limit also get an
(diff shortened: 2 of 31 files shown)


rewitt94:
Questions

1. Whether we need both:
- limit per turn
- optional cumulative limit for longer running sessions

2. How we set a sensible default limit for suspensions per turn (currently 10k)

PR #807, opened 2026-09-03
match CPython for some float rounding behaviour

davidhewitt:
## Summary by cubic
Fixes float-to-integer conversion to match CPython for `round()`, `math.floor()`, `math.ceil()`, `math.trunc()`, and `int()`.

- Replaces i64 clamping with arbitrary-precision promotion for out-of-range floats.
- Adds exact OverflowError/ValueError messages for infinity and NaN.
- Consolidates conversion logic into `LongInt::value_from_f64`.
- Updates `bankers_round` to use the new conversion, fixing `round()` for large floats.
- Removes limitation on mixing float and big-int counts in `Counter`; those operations now produce the correct float result.

<sup>Written for commit 277978fcc9d1833b1da803878156b593f90f3cd5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/807?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty/src/builtins/round.rs
+++ b/crates/monty/src/builtins/round.rs
@@ -101,17 +101,7 @@ pub fn builtin_round(vm: &mut VM<'_>, args: ArgValues) -> RunResult<Value> {
                 // Round to `d` decimal places using banker's rounding.
                 Ok(Value::Float(round_float_to_digits(*f, d)))
             } else {
-                // No digits: round to nearest integer and return int (banker's rounding)
-                if f.is_nan() {
-                    Err(SimpleException::new_msg(ExcType::ValueError, "cannot convert float NaN to integer").into())
-                } else if f.is_infinite() {
-                    Err(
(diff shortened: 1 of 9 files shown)


rewitt94 (approved):
Excellent thanks - this is more correct than clamping and LGTM!

Unrelated question are we going to fix our existing big-int issues?
```
big = int(1e20)          # now exactly 100000000000000000000 (was 9223372036854775807)
float(big)
# TypeError: float() argument must be a string or a real number, not 'int'
```

davidhewitt:
yes all of those should get fixed 👍

PR #810, opened 2026-09-03
don't make static strings a requirement for dump stability

davidhewitt:
bit of a crazy experiment to see if we can reduce merge conflicts

---
## Summary by cubic
Unifies parsing, preparation, compilation, and execution around one dynamically growing `Interns` table, so dump fingerprints no longer depend on the order or set of `StaticStrings` entries. Static strings now serialize by text, module names resolve at runtime, and keyword matching uses executor-independent names; this changes the dump wire format and bumps `DUMP_VERSION` to 10.

**Migration**
- Regenerate dumps created before this change; they will not load.
- A build that does not know a dumped static string preserves its text as an ordinary owned string.

<sup>Written for commit 9f75e7e274e0d074cecc6583f8960e6bb2841aa7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/810?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty/src/bytecode/compiler.rs
```diff
@@ -1080,68 +1079,34 @@ impl<'a> Compiler<'a> {
...
-            self.compile_store(binding)?;
-        } else {
-            // Unknown module - defer error to runtime with RaiseImportError
-            // This allows TYPE_CHECKING imports to compile without error
-            let name_const = self.code.add_const(Value::InternString(module_name))?;
-            self.code.emit_u16(Opcode::RaiseImportError, name_const)?;
-        }
-        Ok(())
+        self.code
+            .emit_u16(Opcode::LoadModule, check_name_index_u16(module_name, position)?)?;
```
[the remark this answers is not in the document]
davidhewitt: I think if we reach the end of the u16 interner range a lot of things will break at the moment. So be it.

crates/monty/src/bytecode/op.rs
```diff
@@ -427,20 +427,11 @@ pub enum Opcode {
     Nop = 105,
 
     // === Module Operations ===
-    /// Load a built-in module onto the stack. Operand: u8 module_id.
+    /// Load a built-in module onto the stack. Operand: u16 module-name StringId.
```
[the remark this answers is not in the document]
davidhewitt: Version is bumped in this PR.

crates/monty/src/bytecode/compiler.rs
```diff
@@ -1080,68 +1079,34 @@ impl<'a> Compiler<'a> {
...
+        self.compile_store(binding)
```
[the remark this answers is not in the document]
davidhewitt: I think if we reach the end of the u16 interner range a lot of things will break at the moment. So be it.

crates/monty/src/args/bind_native.rs
```diff
@@ -567,13 +564,33 @@ impl<C: ContainsHeap> DropWithContext<C> for IterState {
...
+            Self::Ascii(byte) => key.matches(StringId::from_ascii(byte), interns),
+            Self::Static(value) => key.static_string(interns) == Some(value),
+        }
+    }
+}
+
+/// Finds the parameter named by a keyword in declaration order.
+///
+/// Plain positional-only parameters have no keyword identity and do not match.
 fn find_param<'s>(spec: &'s ParamSpec, key: &EitherStr, interns: &Interns) -> Option<(usize, &'s Param)> {
```
rewitt94, line 589: We could reduce the amount of arena look-ups here by calling `let key_static = key.static_string(interns);` once instead of in everything enumeration.

```
fn find_param<'s>(spec: &'s ParamSpec, key: &EitherStr, interns: &Interns) -> Option<(usize, &'s Param)> {
    let key_static = key.static_string(interns);
    spec.params
        .iter()
        .enumerate()
        .find(|(_, param)| {
            param.keyword_name.is_some_and(|name| match name {
                KeywordName::Ascii(byte) => key.matches(StringId::from_ascii(byte), interns),
                KeywordName::Static(value) => key_static == Some(value),
            })
        })
}
```

crates/monty/src/types/str.rs
```diff
@@ -184,12 +184,21 @@ fn ctor_str_arg<'a>(arg: Option<&'a Value>, default: &'a str, vm: &'a VM<'_>) ->
 pub fn allocate_string(s: impl AsRef<str> + Into<Box<str>>, heap: &Heap) -> Value {
     let bytes = s.as_ref().as_bytes();
     match bytes.len() {
-        0 => Value::InternString(StaticStrings::EmptyString.into()),
+        0 => allocate_string_no_interning(s, heap),
```
rewitt94, line 187: `allocate_string("")` no longer returns the canonical empty string. Only some callsites moved to `allocate_string_with_interns`. Others still reach this arm and hand back a fresh heap Str(""), so is empty strings can have different ids.

i.e.
```
empty = ''
b''.decode() is empty      # CPython and main: True, here: False
'{}'.format('') is empty   # CPython and main: True, here: False
f"{''}" is empty           # CPython and main: True, here: False
```

crates/monty/src/types/str.rs
```diff
@@ -184,12 +184,21 @@ fn ctor_str_arg<'a>(arg: Option<&'a Value>, default: &'a str, vm: &'a VM<'_>) ->
...
         1 => Value::InternString(StringId::from_ascii(bytes[0])),
         _ => allocate_string_no_interning(s, heap),
     }
 }
 
+/// Allocates a string while reusing the executor's canonical empty value.
+pub fn allocate_string_with_interns(s: impl AsRef<str> + Into<Box<str>>, heap: &Heap, interns: &Interns) -> Value {
```
rewitt94, line 194: Needing to remember to call this method to get empty string handling seems like a bit of a footgun? Can we combine it with the above somehow (although need for the Interns param seems like a complication)

crates/monty/src/dump_format.rs
rewitt94, line 208: Types also cause some merge conflicts should we implement a similar pattern for them? (As a follow-up I mean ofc)

crates/monty/src/types/str.rs
rewitt94: Maybe we add empty string to the ascii set then create a helper something like this for lookup?

```
pub(crate) fn reserved_string_id(s: &str) -> Option<StringId> {
    match s.as_bytes() {
        [] => Some(EMPTY_STRING_ID),
        [byte] => Some(StringId::from_ascii(*byte)),
        _ => None,
    }
}
```

crates/monty/src/intern.rs
```diff
@@ -1495,41 +1594,69 @@ impl From<Interns> for InternsWire {
...
-        .map(|(index, entry)| {
-            let id = StringId(
-                u32::try_from(INTERN_STRING_ID_OFFSET + index)
-                    .expect("StringId overflow while building reverse interns map"),
-            );
-            (entry.value().clone(), id)
-        })
-        .collect()
+/// Duplicate text is rejected because distinct IDs for equal interned strings
+/// would invalidate the ID-equality fast path used by Python string equality.
```
rewitt94, line 2050: We don't check globally reserved ids here? should we?

rewitt94 (approved):
This is cool! 

Do we have sufficient performance testing to assess this properly - I believe our benchmarks is quite a narrow set of functionality?

davidhewitt:
@codspeedbot explain this regression. Do we have sufficient coverage; should we add some broad benchmarks for this?

crates/monty/src/dump_format.rs
davidhewitt: Yes definitely, all of these should be figured out!

crates/monty/src/intern.rs
```diff
@@ -1495,41 +1979,69 @@ impl From<Interns> for InternsWire {
...
+fn build_string_maps(strings: &StringEntries) -> Result<StringMaps, String> {
+    let mut seen = AHashMap::with_capacity(strings.len());
+    let mut string_id_by_name = AHashMap::new();
+    let static_string_ids = RefCell::new(AHashMap::new());
```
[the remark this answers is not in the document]
davidhewitt: See #873, I don't care

crates/monty/src/value.rs
```diff
@@ -2528,9 +2523,23 @@ impl EitherStr {
...
 ///
 /// Wraps a `StaticStrings` variant to leverage its string conversion capabilities.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub(crate) struct Marker(pub StaticStrings);
 
+impl serde::Serialize for Marker {
+    fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
+        let value: &'static str = self.0.into();
+        serde::Serialize::serialize(value, serializer)
```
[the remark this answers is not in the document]
davidhewitt: Not valid, this PR bumps the dump version.

PR #840, opened 2026-09-10
Store short Python strings inline

art049:
## Changes

Store short Python string contents inline with `CompactString`, already present in the dependency graph.
The inline capacity is 24 UTF-8 bytes on 64-bit targets and 12 on 32-bit targets.
Non-interned strings keep their heap IDs, cached hashes and transparent string serialization.

Convert borrowed and owned inputs directly to compact storage, and encode non-ASCII characters through a stack buffer instead of a temporary `String`.
Generalize the existing `HeapRead` dereference projection so string methods can borrow either inline or allocated contents without copying.
No new unsafe code.

Add snapshot coverage across both inline boundaries, including Unicode, NULs, aliases and dictionary keys.
Also document why `create_page` intentionally returns a `Box`, satisfying the current Clippy lint without changing page allocation.
Use `Duration::from_mins(10)` for the existing 600-second pool test interval to satisfy another current Clippy lint.

## Performance

Local Criterion release build on Apple M5 Pro, 30 samples, 2s warm-up and 5s measurement:

| Benchmark | Before (mean) | After (mean) | Criterion comparison |
| --- | ---: | ---: | --- |
| `list_append_str__monty` | 6.840 ms | 6.287 ms | 8.1% less time; 95% interval: 7.5–8.7% |
| `list_append_int__monty` | 4.198 ms | 4.230 ms | No significant change (p = 0.07) |

Arena layout is unchanged: `HeapData` 72 bytes, `HeapEntry` 96 bytes, `Option<HeapEntry>` 104 bytes.
`Str` grows from 24 to 32 bytes but remains below the existing enum payload ceiling.
These are local wall-time measurements, not a CodSpeed Simulation comparison.

## Validation

- `cargo test -p monty`: 733 passed, 4 ignored.
- `cargo test -p monty --test binary_serde --features memory-model-checks`: 12 passed.
- `cargo test -p monty --lib --features memory-model-checks heap`: 15 passed.
- `cargo run -p monty-datatest -- str__`: 30 Monty/CPython checks passed.
- CLI smoke: 100,000 numeric strings plus ASCII/Unicode/NUL boundary operations; also run against CPython.
- `cargo check -p monty --target wasm32-wasip1`: passed.
- `cargo test -p monty-pool --test pool_test one_huge_write_is_split_at_the_size_threshold`: passed.
- `make format-rs` and `make lint-rs`: passed, including all-features Clippy and import checks.

No API or CPython-behavior change; existing limitations documentation is unchanged.

---
## Summary by cubic
Stores short Python string contents inline with `CompactString`, so small strings avoid heap allocation. Non-interned strings keep their heap IDs, cached hashes, and serialization behavior; no API or CPython-behavior change.

**Performance**
- Local Criterion shows ~8% less time for `list_append_str__monty`; no significant change for `list_append_int__monty`.
- Arena layout is unchanged; `Str` grows from 24 to 32 bytes but stays under the enum payload ceiling.

**Validation**
- Adds round-trip tests covering inline boundaries, Unicode, NULs, aliases, and dictionary keys.
- Memory limit tests now assert only that reported usage exceeds the limit, dropping platform-specific exact byte figures.
- Borrowed and owned inputs convert directly to compact storage without a temporary `String`.
- Documents two Clippy lints: the intentional `Box` return from `create_page` and the `Duration::from_mins` test interval.

<sup>Written for commit 71e8971a4355dc33cb041680c7500cfb3aa16606. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/840?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -757,6 +757,7 @@ dependencies = [
  "castaway",
  "cfg-if",
  "itoa",
+ "serde",
  "static_assertions",
  "zmij",
 ]
@@ -2410,6 +2411,7 @@ version = "0.0.23"
 dependencies = [
  "ahash",
  "chrono",
+ "compact_str",
  "fancy-regex 0.17.0",
  "getrandom 0.3.4",
  "hashbrown 0.16.1",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,6 +99,7 @@ unicode-general-category = "1"
 # or lead CPython's, affecting only recently (re)assigned code points.
 unicode_names2 = "1"
 unicode-normalization = "0.1"
+compact_str = { version = "0.10", features = ["serde"] }
 # Linear-time substring search over `[u8]`, which std only provides for `str`.
 # Already linked transitively via ruff/fancy-regex; this just names it directly.
 memchr = "2.7"
--- a/crates/monty-js/__test__/assertions.ts
+++ b/crates/monty-js/__test__/assertions.ts
@@ -22,24 +22,15 @@ export const t = {
   pass: () => expect(true).toBe(true),
 }
 
-/**
- * Assert a `MemoryError` reports roughly `expected` bytes used against `maxMemory`.
- *
- * The figure is real allocator bytes, so the baseline a session starts from
- * drifts a few dozen bytes between platforms (macOS runs consistently below
- * Linux and Windows) — an exact match would make these tests OS-specific. The
- * tolerance stays far below what a mis-accounted allocation would move it by.
- */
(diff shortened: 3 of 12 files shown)


samuelcolvin:
Looks like a good idea to me. @davidhewitt pls can you review.

Thanks @art049!

PR #842, opened 2026-09-10
Binascii and base64 resource limits checks

rewitt94:
Neither module read the clock or checked a size, so a conversion scanned its whole buffer inside one bytecode instruction — killing the worker rather than raising `TimeoutError`, or exiting the process rather than returning `MemoryError`.

Three kinds of check:

- **Clock polls**, on a 4KiB window so the inner loop keeps its unrolling. Quoted-printable can't be windowed and keeps a per-step test.
- **Size preflights** before allocating: the format's own ratio, or an exact count where a ratio would be wrong. Shrinking conversions need none.
- **Exact reservations** at the size preflighted, appended bytes included — a `Vec` that outgrows its reservation doubles.

Four needed care:

- **`a85decode`** expands, since `z` is a whole zero word, so it counts digits exactly.
- **`b2a_qp`** is sized from what escapes; a flat 3.125 bytes/byte bound rejects plain text.
- **`b2a_base64`** and **`a85encode(adobe=True)`** append to a buffer already filled to the byte, and reserve that too.

### Performance against main
- twelve of sixteen conversions are level 
- `unhexlify` is faster due to removing `collect` at the end of hex_decode.
- `b2a_qp`, `a85decode`, `a2b_qp` all slow - `b2a_qp`, `a85decode` slow from measuring the output before writing it & `a2b_qp` from a per-step clock test it can't window

https://claude.ai/code/session_01Nqjsa2ia5yP6ejfDWqx9WC

---
## Summary by cubic
The `binascii` and `base64` conversions now enforce resource limits while running, so an oversized input raises `TimeoutError` or `MemoryError` instead of killing the worker. Previously both modules skipped clock and memory checks, letting a conversion scan its whole buffer inside one bytecode instruction.

- Clock polls run per 4KiB window; `b2a_qp` and `a2b_qp` poll per step because a soft break can consume a whole line at once.
- Expanding conversions preflight output size before allocating: `a85decode` counts digits exactly (a `z` is a whole zero word), `b2a_qp` counts what escapes, and `b2a_base64` / `a85encode(adobe=True)` reserve the bytes they append.
- Twelve of sixteen conversions stay at parity; `unhexlify` is faster from dropping the trailing `collect`; `b2a_qp`, `a85decode`, and `a2b_qp` are slower.

<sup>Written for commit 7a224aef14ca4793ca79f798a3baaf874e53c911. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/842?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/crates/monty-runtime/tests/subprocess.rs
+++ b/crates/monty-runtime/tests/subprocess.rs
@@ -1199,6 +1199,106 @@ fn loading_a_dump_applies_its_own_memory_limit() {
     );
 }
 
+/// Every `binascii`/`base64` conversion whose output is larger than its input
+/// can be handed a buffer that fits under the soft limit and encodes to one
+/// that clears the allocator's hard ceiling, which kills the worker rather
+/// than raising. Each preflights its output, so the session survives.
+///
+/// The sizes matter: the input must fit under the soft limit while the
+/// expansion clears soft + `BASE_HEADROOM`, which needs a limit above 2 MiB.
+/// `a85decode` is here rather than among the decoders because Ascii85's `z`
+/// stands for a whole zero word, so a buffer of them quadruples.
+#[test]
+fn expanding_codecs_fail_softly() {
+    const LIMIT: u64 = 8 * 1024 * 1024;
+    // smallest expansion here is base64's 4/3, so 5 MB covers every case
+    const SETUP: &str = "import binascii, base64\nsrc = {FILL} * 5_000_000\n";
+    let cases = [
+        ("b'\\xff'", "binascii.hexlify(src)"),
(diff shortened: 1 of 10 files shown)


davidhewitt (commented):
I would like to remove clock poll from native loops (where possible) and rely on grace; can remove those from here and just have the preflight checks.

PR #851, opened 2026-09-12
os module: host-mediated process & identity calls (system, uname, getcwd, cpu_count, getpid)

### LINKED ISSUE #850: os module: host-mediated process & identity calls (system, uname, getcwd, cpu_count, getpid)
Five `os` functions would follow the exact pattern `os.getenv`/`os.environ` already use: the sandbox suspends with a typed `OsFunctionCall`, and the host decides what the answer is.

| Function | Host answer |
|---|---|
| `os.system(command)` | the exit-status int the sandbox should observe |
| `os.uname()` | a 5-field named tuple (sysname, nodename, release, version, machine) |
| `os.getcwd()` | a str |
| `os.cpu_count()` / `os.getpid()` | ints |

**Why**: monty's documented `os` subset omits the process/identity family entirely (limitations/os.md lists `os.system`, `os.getcwd`, `os.uname`, `os.cpu_count`, `os.getpid` under not-implemented), so `import os; os.system("ls")` dies with `AttributeError` before any host policy can apply — while the neighboring filesystem calls (`listdir`, `stat`, `environ`) are already host-mediated os-calls. There's no host-side alternative: `external_lookup` resolves undefined *names*, not module *attributes*, and the sandbox's `os` module is compiled into the worker binary.

**Security**: unchanged. The interpreter performs zero I/O — `os.system` hands the command string to the host and nothing is ever executed by monty itself; that's the same trust model as `Path.write_text`. With no handler, the standard no-handler error applies (current behavior is unchanged for anyone who opts out).

**Details**:
- CPython-exact error wording, including the `posix.` prefix quirk (`posix.uname() takes no arguments (1 given)`) — dual-run test cases
- `os.uname()` returns a named tuple with attribute + index access; repr divergence (`uname_result(...)` vs `posix.uname_result(...)`) matches the existing `StatResult(...)` precedent
- `os.system` accepts str/bytes/PathLike like CPython (bytes decode lossily; documented)
- additive `OsCall` oneof arms (proto tags 25–29); snapshots stay compatible, no dump-version bump

Happy to send the PR.

strawgate:
Closes #850.

Five `os`-module functions that suspend as host-mediated `OsFunctionCall`s, following the exact pattern of `os.getenv`/`os.environ`:

| Function | Host answer |
|---|---|
| `os.system(command)` | the exit-status int the sandbox should observe |
| `os.uname()` | a 5-field named tuple (sysname, nodename, release, version, machine) |
| `os.getcwd()` | a str |
| `os.cpu_count()` / `os.getpid()` | ints |

**The interpreter performs zero I/O** — `os.system` hands the command string to the host and nothing is ever executed by monty itself; the same trust model as `Path.write_text`. With no handler, the standard no-handler error applies, so current behavior is unchanged for anyone who opts out.

Details:
- CPython-exact error wording, including the `posix.` prefix quirk (`posix.uname() takes no arguments (1 given)`) — dual-run test cases pin every message
- `os.uname()` returns a named tuple with attribute + index access; repr diverges like the existing `StatResult(...)` precedent, recorded in limitations/os.md
- `os.system` accepts str/bytes/PathLike (bytes decode lossily; documented)
- additive `OsCall` oneof arms (proto tags 25–29); static-strings fingerprint updated, no dump-version bump
- telemetry spans/metrics for the new calls; the fs/ dispatcher rejects them (no path) so they fall through mounts to the `os` handler like every non-FS call

Follow-ups kept out of this PR: `os.popen` (the capture channel), `os.chdir` + cwd state, the `os.getuid`/`getgid` family.

---
## Summary by cubic
Adds host-mediated `os.system`, `os.uname`, `os.cpu_count`, and `os.getpid`: calls that previously errored now suspend as typed `OsFunctionCall`s and the host supplies the result. `os.getcwd` stays deferred to #828 so the VM can own its working directory; closes #850.

**New Features**
- `os.system` passes `str`, `bytes`, or `os.PathLike` commands to the host, never executing anything itself; the VM's working directory rides along as a kw-only arg.
- `os.uname` returns an attribute- and index-accessible `uname_result`; the other calls return host-provided values.
- Without an `os=` handler these calls error, so `os.system` cannot report a fake success.
- Unknown `OsCall` variants from newer children surface as `os.unknown` and degrade to that no-handler error instead of terminating the session.
- Error wordings match CPython, including the `posix.` prefix on arity errors and `os.system`'s bad-type message.

<sup>Written for commit 9997ac5cf300f8f1d1c2001a3bbe903d35344eb3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/851?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


samuelcolvin:
`getcwd` is already implemented in #828.

samuelcolvin:
I think we need to be super careful about `os.system` and what we do with it.

strawgate:
ai slop comment, feel free to ignore, still looking at this

> 
> Agreed — here's the safety model, made explicit in this push:
> 
> **Fail-closed.** With no handler, `os.system` raises the sandbox's no-handler `RuntimeError` — it never fabricates a success. The variant docstring, `security.md` and `filesystem.md` now all state this in those words.
> 
> **The interpreter is not the executor.** `os.system` hands the command string to the host verbatim and nothing is ever run by monty. If a host handler *chooses* to execute the string it receives, it is acting with the host's own full authority — that's a hole the host author cuts deliberately (the same is already true of every `os` handler, which can do arbitrary things on any call). The docs call this out as "a policy point, not an execution feature".
> 
> **Skew degrades, not dies.** A newer child speaking an unknown arm surfaces as `os.unknown` and degrades to `not_handled` (the child's own no-handler error); the session survives instead of dying on a protocol violation — with unit tests pinning that behavior.
> 
> **On getcwd**: you're right — dropped from this PR; #828's VM-owned working directory is the better design and this branch reserves the wire tag rather than reusing it. `cpu_count`/`getpid`/`uname` remain here; happy to slim further if you'd rather take `system` in a separate round after the identity calls settle — say the word and I'll split it.
>

PR #870, opened 2026-09-15
Charge value export against max_memory so shared graphs cannot OOM the host

samuelcolvin:
Fixes the worker-side finding in HackMonty submission 5Xp8QGM. Follows up #869, which fixed the parent-side decode accounting; this closes the export side.

## The problem

The sandbox heap keeps object identity, but the wire `MontyObject` is a tree with no shared-reference representation, so a value reached through two sibling edges is exported twice. A tiny shared graph expands exponentially:

```python
x = [0]
for _ in range(20):
    x = [x]
for _ in range(15):
    x = [x, x]
x
```

That is 36 heap objects but 753,663 wire nodes. The conversion ran after the execution clock stopped and charged nothing against `max_memory`, so a ~3 MB event retained ~250 MB in the parent, bounded only by the worker's hard allocator ceiling.

## The fix

`MontyObject::from_value` now charges each exported node against the session's `max_memory` as the tree is built (`vm.heap.tracker.check_allocation`), the same soft-limit mechanism the rest of the interpreter uses. Once the budget is spent it stops recursing and the top-level conversion surfaces the `MemoryError`, so an over-budget export is refused gracefully and the worker and session survive rather than the process dying.

To carry the error out, the conversion is now fallible: `from_value_inner` threads a memory accumulator, and `new` / `from_value` / the call-argument builders (`ArgValues::into_py_objects`, `KwargsValues::into_py_objects`) and `os.getenv`'s default export return `RunResult`. Recursion-depth overflow still degrades to its `<deeply nested>` placeholder unchanged; only memory exhaustion becomes an error.

## Tests

`crates/monty-runtime/tests/subprocess.rs`:
- `exporting_a_shared_graph_is_charged_against_max_memory`: the report's program returns `MemoryError` under a 4 MiB limit and the worker survives to run the next feed.
- `exporting_a_small_shared_graph_still_succeeds`: the same shape at depth 3 round-trips intact, so the budget changed no output for graphs that fit.

Full `monty`, subprocess, pool, ref-count-return and 1220 datatest cases pass.

## Docs

`limitations/resource_limits.md` and `docs/resource-limits.md`.

## Out of scope

Counting conversion time against `max_duration_secs` (the pool turn timeout already backstops it), and the event-side repeated-field audit from the #869 review (`Print.segments`, `RaisedException.traceback`), which is its own change.

https://claude.ai/code/session_01N9w1kUnJsNpPqgwNEktLRp

---
## Summary by cubic
Charges sandbox value export against `max_memory` as it builds the host-side tree, preventing heavily shared graphs from expanding without bound. Previously, returned values, host-call arguments, and `os.getenv` defaults were exported without accounting; they now raise `MemoryError` when the budget is exhausted, while in-budget output and recursion-depth placeholders remain unchanged and the worker session survives.

**Tests and docs**

- Adds coverage for oversized shared graphs, successful small graphs, and worker reuse after a rejected export.
- Documents the host-side tree conversion and its `max_memory` behavior.

<sup>Written for commit 74c57f840d958f46a11a965bb3a84f7ec6b25254. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/870?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


samuelcolvin:
replaced by #871.

PR #875, opened 2026-09-16
Fix async task scheduling for batched host failures

flyingmutant:
When trying to update to the current Monty, our tests have caught a possible scheduler stall in the case of several failing tasks in one batch.

This is my attempt to fix it properly – scheduler is a bit simpler overall as a result (it delivers exceptions when switching to failing task instead of immediately).

---
## Summary by cubic
Previously, host rejections in one result batch were delivered immediately and could stall the scheduler. They are now queued like successful wakeups, so each task raises its own exception and runs `except`/`finally` before an unhandled error fails its gather.

**Bug Fixes**
- Removed the immediate failure-delivery shortcut; failing tasks now go through the same activation path as successful ones.
- Released exception stacks when tasks exit so teardown doesn't leave stale exception context.

**Migration**
- `DUMP_VERSION` is bumped because the serialized task state changed; old dumped sessions won't load.

<sup>Written for commit b1ba569020e8808a4183d5512161b13d73a6b523. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/875?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


davidhewitt (approved):
Thanks, this is indeed a nice simplification 👍

PR #885, opened 2026-09-17
expose otel trace context from snapshots

davidhewitt:
## Summary by cubic
Adds `traceContext()` (JS) / `trace_context()` (Python) to all snapshot types so manual handlers can nest host tracing under the suspension's span when Monty instrumentation is enabled.

- Returns the OTel Context captured at `feedStart` / `feed_start` (or `loadSnapshot` / `load_snapshot` when restoring), with the suspension's span set when Monty tracing is active; composition failures fall back to the captured context, with JS reporting the failure via `diag.warn`.
- Without Monty tracing, returns the captured context unchanged.
- Context is never serialized; restoring snapshots captures the restoring caller's context.
- Calling after resume throws; contexts obtained earlier remain usable but don't keep the span open.
- Python requires `opentelemetry-api`; calling without it raises `ImportError`.
- `resumeAuto()` / `resume_auto()` already activate the suspension span around callbacks.

<sup>Written for commit 0a6cc72b0026b44b0a789944387f319d2f6aaf5c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/pydantic/monty/pull/885?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


crates/monty-python/python/pydantic_monty/_monty.pyi
```diff
@@ -3,6 +3,7 @@ from collections.abc import Mapping
 from pathlib import Path
 from typing import Any, Callable, Literal, NoReturn, final
 
+from opentelemetry.context import Context
```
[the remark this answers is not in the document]
davidhewitt: I believe this to be ok; mypy will suppress the errors from this package if it's a dependency of the actual user project.

crates/monty-js/README.md
```diff
@@ -259,6 +259,32 @@ while (!(snap instanceof MontyComplete)) {
 console.log(snap.output) // 'hello Ada!'
 ```
 
+For manual handlers, all three snapshot types expose `traceContext()`, returning an OpenTelemetry `Context`.
```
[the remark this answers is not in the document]
davidhewitt: This example is duplicated into the docs.

crates/monty-python/src/telemetry.rs
```diff
@@ -148,6 +142,57 @@ pub(crate) fn callback_context(py: Python<'_>, context: &Context) -> Option<Py<P
         .map(Bound::unbind)
 }
 
+/// Captures feed/load context without requiring OpenTelemetry for ordinary execution.
+pub(crate) fn capture_otel_context(py: Python<'_>) -> Option<Py<PyAny>> {
+    py.import("opentelemetry.context")
```
rewitt94, line 147: We do full import on every feed? Maybe we could cache somehow?

crates/monty-js/ts/callbackContext.ts
```diff
@@ -17,6 +20,16 @@ export function bindPrintCallback(callback: NativePrintCallback): NativePrintCal
   return handlers?.bindPrint(callback) ?? callback
 }
 
+/** Returns the suspension's span in its captured context, preserving baggage and other entries. */
+export function getCallbackContext(parent: string | undefined, captured: Context): Context {
+  try {
+    const span = handlers?.span(parent)
+    return span === undefined ? captured : trace.setSpan(captured, span)
+  } catch {
+    return captured
```
rewitt94, line 30: In what event to does this fail ? Can/should we differentiate the feature being disabled and an error?

crates/monty-js/ts/callbackContext.ts
```diff
@@ -1,10 +1,13 @@
-// Node installs these handlers; browser transports use the direct callback path.
+// Node supplies suspension spans; browser transports preserve the captured context.
```
rewitt94, line 1: > browser transports preserve the captured context

Do we want a browser test for this?

rewitt94 (approved):
Overall this looks fine to me 

Claude identifies there as some divergences in the js/python implementations

```
Divergences

- Span lookup — Python reads it live through the checkout mutex on every call; JS captures the span key onto the snapshot at construction.
- Failure modes — Python's accessor can fail ('snapshot session is busy', 'snapshot session is closed'); JS's cannot.
- OTel dependency — optional in Python, so a missing opentelemetry-api raises ImportError; a hard dependency in JS, which always degrades quietly instead.
- Source of truth for the span — Python has two (a contextvar for callbacks, the bridge registry for manual handlers); JS has one, shared by both paths.
- Type surface — Python repeats the method across six sync + async snapshot types; JS has three classes sharing one SingleUse base.
- Error types — RuntimeError / ImportError in Python; plain Error with a message in JS.
```

Left a couple other nits too - 
but nothing blocking

crates/monty-python/src/telemetry.rs
davidhewitt: For now will leave as-is, import system should hopefully cache (though I agree this is potentially suboptimal).

